PR #41, opened 2014-12-22
Add Expiring Token to Endpoint

### LINKED ISSUE #39: Add Expiring Token to Endpoint for DDOS Protection
Add an expiring, random or encrypted token to webmention endpoints, preventing the accumulation of lists of endpoints and forcing attackers to look up the webmention endpoint of each of the sites they want to use to DDOS a victim.

http://indiewebcamp.com/DDOS#Expiring_token_in_endpoint

peterwilsoncc:
Add a rolling token to the webmention endpoint. 

A new token is generated every twelve hours, the current and previous two tokens are valid at any one time.

Native WP nonces are out as they include the session and user IDs.

Fixes #39

### DIFF
--- a/webmention.php
+++ b/webmention.php
@@ -70,8 +70,8 @@ class WebMentionPlugin {
   }
 
   /**
-   * get the three valid endpoint codes. In the returned array,
-   * the first item -index 0- is the most correct
+   * generate a valid expire code. 
+   * Three possible values are valid at any one time, ticks: 0, 1, or 2
    * 
    * @return array
    */
--- a/webmention.php
+++ b/webmention.php
@@ -94,7 +94,8 @@ class WebMentionPlugin {
     // as above, always use the lgoged out token.
     $token = '';
 
-
+    // custom hash used rather than standard nonce to prevent session data polluting the
+    // web mention endpoint. The endpoint needs to remain the same for all users.
     $expire_code = wp_hash( $expire_code . '|' . $action . '|' . $uid . '|' . $token, 'nonce' );
     
     return $expire_code;
--- a/webmention.php
+++ b/webmention.php
@@ -75,15 +75,17 @@ class WebMentionPlugin {
    * 
    * @return array
    */
-  public static function expire_codes() {
+  public static function expire_code( $tick = 0 ) {
     $action = 'web mention endpoint';
     $time_format = 'Y-m-d a';
     $time_block = 12 * HOUR_IN_SECONDS;
-    $valid_codes = array(
-      date( $time_format, time() ),
-      date( $time_format, time() - $time_block ),
-      date( $time_format, time() - ( 2 * $time_block ) )
-    );
+    $tick = abs( intval( $tick ) );
+    if ( 3 < $tick ) {
+        // something wrong, tick too high/
+        // use default
+        $tick = 0;
+    }
+    $expire_code = date( $time_format, time() - ( $tick * $time_block ) );
     
     // always use logged out user code, endpoint may be looked up by a logged in user
     // while the web mention comes from a logged out user (using curl or similar)
@@ -94,11 +96,9 @@ class WebMentionPlugin {
     $token = '';
 
 
-    foreach ( $valid_codes as $key => $expire_code ) {
-      $valid_codes[$key] = wp_hash( $expire_code . '|' . $action . '|' . $uid . '|' . $token, 'nonce' );
-    }
+    $expire_code = wp_hash( $expire_code . '|' . $action . '|' . $uid . '|' . $token, 'nonce' );
     
-    return $valid_codes;
+    return $expire_code;
   }
   
   
@@ -116,13 +116,13 @@ class WebMentionPlugin {
     }
     else {
       // check if the end point has expired
-      $valid_endpoint_codes = WebMentionPlugin::expire_codes();
+      $valid_ticks = array( 0, -1, -2 );
       
       $supplied_code = get_query_var( 'webmention' );
       $is_valid = false;
  
-      foreach ( $valid_endpoint_codes as $expire_code ) {
-        if ( hash_equals( $expire_code, $supplied_code ) ) {
+      foreach ( $valid_ticks as $tick ) {
+        if ( hash_equals( WebMentionPlugin::expire_code( $tick ), $supplied_code ) ) {
           $is_valid = true;
           break;
         }
@@ -611,8 +611,7 @@ class WebMentionPlugin {
    */
   public static function html_header() {
     // backwards compatibility with v0.1
-    $valid_endpoint_codes = WebMentionPlugin::expire_codes();
-    $endpoint_code = $valid_endpoint_codes[0];
+    $endpoint_code = WebMentionPlugin::expire_code();
     echo '<link rel="http://webmention.org/" href="'.site_url("?webmention=" . $endpoint_code ).'" />'."\n";
     echo '<link rel="webmention" href="'.site_url("?webmention=" . $endpoint_code ).'" />'."\n";
   }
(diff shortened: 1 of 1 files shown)


pfefferle:
Hey @peterwilsoncc thanks for the changes and sorry for the late response (I am a bit busy at the moment)

I had a look at your code and was curious if it is also possible to add the feature with native nonces? http://codex.wordpress.org/WordPress_Nonces

webmention.php
```diff
@@ -81,13 +114,34 @@ public static function parse_query($wp) {
...
+        }
+      }
+    }
 
     // plain text header
     header('Content-Type: text/plain; charset=' . get_option('blog_charset'));
 
+    // fail if invalide endpoint
+    if ( false == $is_valid ) {
+      status_header(400);
```
voxpelli: A `403` would probably be better here, http://httpstatus.es/403, as the server understands the request, but refuses to handle it.

A `400` is more of a generic error where the server don't really understand what the client actually wanted to do: http://httpstatus.es/400

pfefferle: makes sense! :+1:

webmention.php
```diff
@@ -70,6 +70,39 @@ public static function query_var($vars) {
...
+        // something wrong, tick too high/
+        // use default
+        $tick = 0;
+    }
+    $expire_code = date( $time_format, time() - ( $tick * $time_block ) );
+    
+    // always use logged out user code, endpoint may be looked up by a logged in user
+    // while the web mention comes from a logged out user (using curl or similar)
+    $uid = 0;
+    $uid = apply_filters( 'nonce_user_logged_out', $uid, $action );
```
voxpelli: Looking into what this filter actually does it seems like this filter is meant to enable other plugins to base the nonce on some _session_ data:https://core.trac.wordpress.org/changeset/21837/trunk As WebMention requests should be curlable and thus stateless I'm of the impression that this filter would not give the desired effect.

webmention.php
```diff
@@ -81,13 +114,34 @@ public static function parse_query($wp) {
     if (!array_key_exists('webmention', $wp->query_vars)) {
       return;
     }
-
-    $content = file_get_contents('php://input');
-    parse_str($content);
+    else {
+      // check if the end point has expired
+      $valid_ticks = array( 0, -1, -2 );
```
voxpelli: Why are numbers negative here? And why have an array of numbers combined with a `foreach` rather than just a regular `$length = 3; for ($i = 0; $i < $length; $i++) {}`?

webmention.php
pfefferle: The native nonce support also calls this filter, so I think it would be nicer to use the native feature and filter the nonce to skip "session" like support and to only add some kind of a stateless nonce.

webmention.php
pfefferle: My suggestion would be:

* add a setting to enable/disable nonce support
* using `$nonce = wp_create_nonce( 'my-action_'.$post->ID );` so it is possible to use all other wp_nonce native features
* using the filter to return a "stateless" UID

What are you thinking?

webmention.php
```diff
@@ -70,6 +70,39 @@ public static function query_var($vars) {
...
+    $action = 'web mention endpoint';
+    $time_format = 'Y-m-d a';
+    $time_block = 12 * HOUR_IN_SECONDS;
+    $tick = abs( intval( $tick ) );
+    if ( 3 < $tick ) {
+        // something wrong, tick too high/
+        // use default
+        $tick = 0;
+    }
+    $expire_code = date( $time_format, time() - ( $tick * $time_block ) );
```
voxpelli: Any reason to not use `wp_nonce_tick()` instead of calculating the tick from scratch? Not sure what's the best – but perhaps add a comment with a reason for why you chose what you did would be good?

webmention.php
```diff
@@ -70,6 +70,39 @@ public static function query_var($vars) {
...
+    // always use logged out user code, endpoint may be looked up by a logged in user
+    // while the web mention comes from a logged out user (using curl or similar)
+    $uid = 0;
+    $uid = apply_filters( 'nonce_user_logged_out', $uid, $action );
+    
+    // as above, always use the lgoged out token.
+    $token = '';
+
+
+    $expire_code = wp_hash( $expire_code . '|' . $action . '|' . $uid . '|' . $token, 'nonce' );
```
voxpelli: This looks very similar to the code in `wp_create_nonce()`: https://core.trac.wordpress.org/browser/tags/4.1/src/wp-includes/pluggable.php#L1758 But without the substr()? Any reason for that? Should perhaps add a comment on why it's not possible to use `wp_create_nonce()` (I guess the reason is the risk of getting session-dependent data which would stop the endpoint definition from being stateless and "curlable")

Ideally, long term, a patch to WordPress core that would allow a stateless `wp_create_nonce()` to be used out of the box would be great as well I think :) This is a great use case for such a patch.

pfefferle: Ah ok, here is the problem https://core.trac.wordpress.org/browser/tags/4.1/src/wp-includes/pluggable.php#L1750 the logged out code is only called if there is no session! That is indeed a problem!

webmention.php
voxpelli: I like the thinking, but the `` from https://core.trac.wordpress.org/browser/tags/4.1/src/wp-includes/pluggable.php#L1755 makes it impossible as far as I can see to make the built in `wp_create_nonce()` stateless as far as I can see. `'nonce_user_logged_out'` doesn't solve the stateless part – it was intended to increase statefulness rather than decreasing it so as long as there's a session with a token it will be impossible to make `wp_create_nonce()` stateless even with it.

voxpelli:
All in all: I'm a bit concerned with how a client should be able to know when an endpoint will become invalid. Imaging a big social network implementing WebMentions – you don't want them to fetch a post before every new mention of that post.

But how often should it fetch a post? Should it keep going with an endpoint until it receives a 403 and then fetch again? Should there be a cache-header set to the length of time of the token life so the client can check for that to know how long to cache the site? If the cache is set to longer than the token life, can't that make it hard for a well-behaved client to actually receive a new token? So would this effectively make it so that no blog post can be cached for longer than `$time_block = 12 * HOUR_IN_SECONDS;` x 3?

I guess this is more of an IndieWeb / WebMention question than a PR question but still feel it kinds of needs to be addressed.

voxpelli:
Thinking about this some more: I would advise against including this as a default in a plugin like this for now – I'm not sure this route is the best way to go down to prevent DDoS attacks and it adds a lot of possibly complexity on the client side when it comes to caching etc which is not in favor of WebMention adoption etc at the moment in my opinion.

What I do however think is that it should be easy to dogfeed a mechanism like this so that it can easily be tried out.

So I would suggest adding a filter for the value of the `?webmention=` query param so that another plugin can add a nonce in there rather than having it be the default `endpoint`, as suggested by this PR, and then add an action/filter (don't remember which is the most appropriate, was a while since I was an active WP plugin writer) in the beginning of `parse_query()` so that the same plugin could validate that the defined nonce was correct.

Would such a solution work for everyone? I hope I'm not sounding too harsh stumbling in here from nowhere with lots of opinions over a month after the PR was sent in.

peterwilsoncc:
Finally have a chance to reply to these.

Brid.gy caches end points for up to 24 hours so I am I'm using custom nonces to ensure they expire between 24 & 36 hours, current & two previous ticks. Native nonces expire between 12 & 24 hours and use session data.

Line comments I'll answer inline.

P

webmention.php
peterwilsoncc: Y, update will be pushed shortly.

webmention.php
peterwilsoncc: Native nonces don't allow skipping of the session data. 

Makes sense to remove this filter, will push.

webmention.php
peterwilsoncc: WP native uses negative numbers when generating the tick, I wanted to match.

Per voxpelli on session data.

webmention.php
peterwilsoncc: nonce_life is filterable in wp_tick & I wanted to ensure it was kept @ twelve hours. 

Would like to know your thoughts on this @pfefferle @voxpelli 

P

webmention.php
peterwilsoncc: Comment added, pushing shortly.

aaronpk:
fwiw I just removed this functionality from my own webmention endpoint, as I wasn't seeing any immediate benefit. I think there are likely other better solutions to (D)DOS attacks, and this just feels like an unnecessary complication.

peterwilsoncc:
I did the same when re-launching recently, for much the same reason.

I'll close this off: conflicts and needs a rethink.

pfefferle:
BTW: I changed some hooks and added new ones, to enable these kinds of security features as third party plugins.

PR #43, opened 2015-05-12
How to enable WebMentions without coding

killerog:
The current description didn't make it really clear to me that this was an option without coding stuff into the WordPress core or maybe to the WPCron (does that still exist btw? haven't worked with WP for years).


pfefferle:
Thanks!

PR #44, opened 2015-05-12
Text edit ended up on the same line

killerog:
Sorry, not very familiar with this markdown syntax :-)


killerog:
I have no grunt here :) Should I change the txt with an extra linefeed too, or will that generate the right way?

pfefferle:
It would be nice if you can add the line in the txt file too. the readme.txt is the base to render the markdown file.

killerog:
Sure, got nothing better to do anyway in the breaks at a conference ;-)

PR #47, opened 2015-06-07
replaced link matching regex with http://daringfireball.net/2010/07/i…

petermolnar:
added regex from http://daringfireball.net/2010/07/improved_regex_for_matching_urls to match all urls ( in-text & markdown ); tested, working

since it's kind of specific and may not fit everyone's needs I just wanted to add as a possibility.

### DIFF
--- a/webmention.php
+++ b/webmention.php
@@ -408,8 +408,15 @@ class WebMentionPlugin {
     $links = array();
 
     // Find all external links in the source
-    if (preg_match_all("/<a[^>]+href=.(https?:\/\/[^'\"]+)/i", $post->post_content, $matches)) {
-      $links = $matches[1];
+    // http://daringfireball.net/2010/07/improved_regex_for_matching_urls
+    if ( preg_match_all('/(?xi)\b((?:https?:\/\/|www\d{0,3}[.]|[a-z0-9.\-]+[.][a-z]{2,4}\/)(?:[^\s()<>]+|\(([^\s()<>]+|(\([^\s()<>]+\)))*\))+(?:\(([^\s()<>]+|(\([^\s()<>]+\)))*\)|[^\s`!()\[\]{};:\'".,<>?«»“”‘’]))/', $post->post_content, $matches )) {
+
+        // resolve issues with [url](url) type markdown links
+        foreach ($matches[2] as $cntr => $url )
+            if ( !empty($url) )
+                $matches[1][$cntr] = $url;
+
+        $links = $matches[1];
     }
 
     // filter links
--- a/webmention.php
+++ b/webmention.php
@@ -407,17 +407,20 @@ class WebMentionPlugin {
     // initialize links array
     $links = array();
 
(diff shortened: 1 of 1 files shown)


pfefferle:
First of all: Thanks for the pull request :)

...but I have tried the code with a demo MD text and it seems to break when using something like:

```Markdown
[https://github.com](https://github.com)
```

pfefferle:
What do you think about adding something like:

```PHP
preg_match_all("/\[([^\]]+)\]\((https?:\/\/[^)]+)\)/i", $post->post_content, $matches )
```

as a second check?

petermolnar:
That is weird, it is working for me, though I'm under PHP 5.6.8 if that matters.

The foreach is there explicitly to sort the [https://github.com](https://github.com) trouble because in those cases the match goes into $matches[2] and not into [1] so I need to override that.
I'd fix the regex, but this is black magic level and way beyond me.

petermolnar:
I'll look into this deeper.

pfefferle:
Can you post a link to a gist with your demo text here? So that we could test it with the same data?

pfefferle:
I used 5.5.20 btw.

petermolnar:
I've made a few tests and the daingfireball regex seems to be overreacting in the (url)[url] case with a few hundred thousand backtracks.
So I'm currently checking 
```
if (preg_match_all("/\b(?:http|https)\:\/\/?[a-zA-Z0-9\.\/\?\:@\-_=#]+\.[a-zA-Z0-9\.\/\?\:@\-_=#]*/i", $post->post_content, $matches)) {
        $links = $matches[0];
    }
```
which I put together from at least 3 different patterns.

The text I'm using now:
```
[https://domain.com/test](https://domain.com/test)

[Link to test 1](https://domain.com/test-1)

https://domain.com/test-2

<a href="https://domain.com/test-3">https://domain.com/test-3</a>

<a href="https://domain.com/test-3">Link to test-3</a>
```

so far it matches every url, so an array_unique is still needed.

pfefferle:
The question is, can we ping every URL or do we have to skip things like:

```
https://domain.com/test-2
```

because they might be text-only and not (hyper)linked to the URL.

petermolnar:
That's the thing; since I'm writing my WP posts in Markdown ( and I store them in md as well ), I'm expecting any urls to be picked up, since they are all converted to links.

Also, there is the make_clickable function in WP which I don't know if it's by default applied, but if yes, then everyone is expecting all urls to become links, thus pingbacks, in my opinion.

pfefferle:
Good point, will check the `make_clickable` thing later.

pfefferle:
This is also problematic, because it does not support Umlauts in the domain name or the path.

pfefferle:
Why have you closed it? I think this is still a good idea!

petermolnar:
Because it messes up things too much; I've ended up using your already in place filter to re-parse my posts which solved all my issues.

PR #48, opened 2015-06-18
Support Comment Notifications for Webmention

dshanske:
This should address the issue.


pfefferle:
should we replace all notifications or only the WebMention (or linkback) ones?

dshanske:
This code is replacing only the webmention type. I also sent a PR for Semantic Linkbacks with expanded code.

dshanske:
I'm closing this for now. I've realized the problem in Semantic Linkbacks needs to be addressed first in webmentions. Going back to the drawing board.

PR #49, opened 2015-06-25
Webmention Aware Notifications

dshanske:
This adds the webmention type as a supported notifications type. All the text is the same as in default WordPress. Much cleaner than before.


dshanske:
Any news on this Pull?

pfefferle:
I was at vacation, sorry... Will have a look at it on monday...

dshanske:
Didn't realize you were on vacation.

pfefferle:
I have checked this and really can't find any advantages against the core implementation. Can you explain me the advantages?

dshanske:
The core implementation treats webmentions like comments instead of like pingbacks. This has them treated like pingbacks.

dshanske:
I'm closing this pull request in the hopes of resolving my Ticket with WordPress to make this less invasive to write. 

In the interim, I think I may, as inelegant as it is, create a plugin with a pluggable version of the notification functions that does know about webmentions.

PR #50, opened 2015-07-28
adding kses

petermolnar:
wp_kses is a function to sanitize HTML. I've added wp_kses_post to the remotely pulled contents to prevent things like XSS, hidden iframes, etc.


pfefferle:
Thanks

pfefferle:
Thanks

PR #55, opened 2015-10-15
Do not unapprove already-approved webmention

singpolyma:
If we're updating an existing mention, keep the existing approval
status.


pfefferle:
Thanks @singpolyma!

PR #58, opened 2015-10-15
Comments sent via salmon are also dupes

singpolyma:
Or anyone else who can't use comment_author_url as the original link,
but can use a _crossposting_link meta value.

### DIFF
--- a/webmention.php
+++ b/webmention.php
@@ -230,7 +230,14 @@ class WebMentionPlugin {
 		if ( ! empty( $comments ) ) {
 			$comment = $comments[0];
 		} else {
-			$comment = null;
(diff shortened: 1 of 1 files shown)


singpolyma:
I'm going to try changing the bottom of my default_request_handler to this:

```
                if ( $comment_ID ) {
                        // set header
                        status_header( apply_filters( 'webmention_success_header', 200 ) );

                        // render a simple and customizable text output
                        echo apply_filters( 'webmention_success_message', get_comment_link( $comment_ID ) );

                        do_action( 'webmention_post', $comment_ID );
                } else {
                        status_header( 500 );
                        $wpdb->print_error();
                }
```

and see what happens

PR #81, opened 2016-06-06
Replace with built-in WordPress functions

dshanske:
This replaces make_absolute_url with the built in version of same, and replaces a pure instance of parse_url, with wp_parse_url...a wrapper which supports schemeless URLs on earlier versions of PHP

### DIFF
--- a/class-webmention-sender.php
+++ b/class-webmention-sender.php
@@ -206,7 +206,7 @@ class Webmention_Sender {
 	 */
 	public static function discover_endpoint( $url ) {
 		/** @todo Should use Filter Extension or custom preg_match instead. */
-		$parsed_url = parse_url( $url );
+		$parsed_url = wp_parse_url( $url );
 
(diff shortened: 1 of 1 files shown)


pfefferle:
isn't it `make_absolute_url` instead?

dshanske:
Correct, it is late, and I am stuck in the airport waiting to get home. I shouldn't have tried to write code this late.

pfefferle:
No problem, I am sorry that I forgot to merge and that you are that patient and persistent!

pfefferle:
The `WP_Http` is a very nice change BTW (didn't know about the `make_absolute_url` before). The less code the better ;)

dshanske:
I hope to make a lot of positive improvements in the codebase. If you skimmed the experimental version, I'm playing with a lot of different ideas and testing them before I write the pull requests. The ones I plan on sending first are things like this that remove code.

pfefferle:
If you update the pull request, I will merge it.

dshanske:
Updated

pfefferle:
thanks :)

PR #82, opened 2016-06-07
Replace Direct Calls to Database with WordPress Functions

dshanske:
This bumps the version requirement to 4.5, but it avoids any direct calls to the database.


pfefferle:
Merged by hand

PR #84, opened 2016-06-09
Unify HTTP Arguments, Set Minimum Version, and Cleanup

dshanske:
HTTP requests now use the same arguments for consistency, started changelog, and updated README with requirement for 4.5.


pfefferle:
merged by hand

PR #85, opened 2016-06-11
Target Before Source Verification and Related Changes

dshanske:
This swaps the target verification before the source. To assist in that, it starts creating the $commentdata array and passes that, as it will be needed anyway.

The default request handler is now responsible for source verification and content creation...which means it could be replaced with an async handler as needed.

'vouch' is now a parameter passed into the comment array, which allows for vouch testing support at the preprocess_comment level or a handler that will handle vouch to be used. 

To match changes made to pingbacks in WordPress, it now passes variables remote_source and remote_source original into the commentdata. This will allow a future update to Semantic Linkbacks that will avoid it having to pull the source an additional time. 

Also fixed an error in the latest refactoring that generated a strict standards warning because an init function didn't have static.

And, as the code is parsing meta tags, it will look for a title in the meta tags before parsing the title.

This time I did do some testing to make sure it worked as expected.

### DIFF
--- a/includes/class-webmention-receiver.php
+++ b/includes/class-webmention-receiver.php
@@ -9,7 +9,7 @@ class Webmention_Receiver {
 	 * Initialize the plugin, registering WordPress hooks
 	 */
 	public static function init() {
-		add_filter( 'query_vars', array( 'Webmention_Receiver', 'query_var' ) );
+		add_filter( 'query_vars', array( 'Webmention_Receiver', 'query_vars' ) );
 		add_action( 'parse_query', array( 'Webmention_Receiver', 'parse_query' ) );
 
 		add_action( 'admin_comment_types_dropdown', array( 'Webmention_Receiver', 'comment_types_dropdown' ) );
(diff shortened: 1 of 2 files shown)


pfefferle:
For your interest: I am working on some further changes of your pull request, to keep it as simple as possible. I like the refactoring and I am only removing some filters and small restructorings. I removed the Vouch and CSRF stuff completely until the spec is final ( CSRF is a GET param btw. ;) ).

I hope I will be able to commit this weekend, to have a look at the SL changes. I love the `preprocess_comment` thingy a lot... good work! hopefully this will make it also into one of the next versions https://core.trac.wordpress.org/ticket/36302

dshanske:
This was the biggest change in the set so far. There was no way to break it up. To switch to target first per spec it had to change to that degree. As for Vouch, as I said, I wanted a way to pass it so a second plugin could have easily experimental support. As for the ticket, we are getting $commentdata passed to edit_comment in 4.6 at least.

pfefferle:
As I said, I love the changes... but for the Couch and CSRF we have to use GET instead of post... I removed it for now, but we can re-add it if I have merged my and your changes.

PR #88, opened 2016-06-25
bug fix for check that target is on site

snarfed:
get_site_url() includes the wordpress directory, so we want home_url() instead. also don't keep going if we hit this. :P


singpolyma:
+1 I just got bit by this bug.

PR #90, opened 2016-07-14
Move Hook so Triggers on All Attempts to Send Webmentions

dshanske:
To add the ability to log errors and successes if desired.


pfefferle:
thanks

PR #92, opened 2016-08-03
Refactor and Move Filters Around

dshanske:
Tried to continue this work.

### DIFF
--- a/includes/class-webmention-receiver.php
+++ b/includes/class-webmention-receiver.php
@@ -23,8 +23,6 @@ class Webmention_Receiver {
 
 		// default handlers
 		add_filter( 'webmention_comment_data', array( 'Webmention_Receiver', 'check_dupes' ), 1, 1 );
-		add_filter( 'webmention_comment_data', array( 'Webmention_Receiver', 'default_title_filter' ), 9, 1 );
-		add_filter( 'webmention_comment_data', array( 'Webmention_Receiver', 'default_content_filter' ), 10, 1 );
 		add_action( 'webmention_request', array( 'Webmention_Receiver', 'default_request_handler' ), 10 );
 	}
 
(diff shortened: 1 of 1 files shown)


includes/class-webmention-receiver.php
```diff
@@ -183,19 +189,23 @@ public static function default_request_handler( $data ) {
 		$commentdata = compact( 'comment_author', 'comment_author_email', 'comment_content', 'comment_parent', 'remote_source', 'remote_source_original', 'comment_approved', 'comment_type' );
 		$commentdata = array_merge( $commentdata, $data );
 
-		$commentdata = apply_filters( 'webmention_comment_data', $commentdata );
+		// Add Default Content which can be overridden using preprocess_comment
+		$commentdata = self::default_content( $commentdata );
+		$commentdata = self::default_title( $commentdata );
```
dshanske: Instead of having a filter for the content and title this can be done by preprocess_comment, so the default title should just be set, and not be a filter from its own. If we want to bring it back with a...to be deprecated notice, we can put it inside the return for the title and content functions.

includes/class-webmention-receiver.php
```diff
@@ -205,9 +215,9 @@ public static function default_request_handler( $data ) {
 		status_header( apply_filters( 'webmention_success_header', 200 ) );
```
dshanske: With the standard stable, do we need this filter? Any scenario that wouldn't return a 200 on success would require a different request handler.

includes/class-webmention-receiver.php
```diff
@@ -283,8 +292,8 @@ public static function check_dupes( $commentdata ) {
 			$commentdata['comment_ID'] = $comment->comment_ID;
 			$commentdata['comment_approved'] = $comment->comment_approved;
 		}
-
-		return $commentdata;
+		// Add hook for Semantic Linkbacks to Add Duplicate Comment Data
+		return apply_filters( 'webmention_check_dupes', $commentdata );
 	}
```
dshanske: The idea of this is that for webmentions, Semantic Linkbacks can link directly into the duplicate webmention filter. There is no point in checking this for Pingbacks and Trackbacks, because they both stop if they find a duplicate. It is part of their specification.

dshanske:
In light of the request to use the API code. Going to close this until that is resolved.

Issue #93, opened 2016-08-03
Webmentions not sent for custom post types


singpolyma:
The issue seems to be that `get_posts( array( 'meta_key' => '_mentionme', 'post_type' => 'any', 'fields' => 'ids', 'nopaging' => true ) )` does not return custom posts.  Even though it says `any` there.

dshanske:
WordPress required unit tests for WP_Query, so it should reveal if any stopped working. So I am stumped.

singpolyma:
When I changed it to say  `array('post', 'actionstream')` it works, but with `any` it does not.

dshanske:
Similar: https://core.trac.wordpress.org/ticket/36824

dshanske:
I want to check out the unit tests for that feature.

dshanske:
I am going to write some tests of my own.

dshanske:
The post type actionstream, is it from a public plugin or something private?

singpolyma:
@dshanske from the diso-actionstream plugin (though I have some local hacks to that for sure).

dshanske:
Did it work when we did a direct database query?

singpolyma:
@dshanske it used to work

dshanske:
So, I need to test this. Will try to figure it out. If it is a Core problem, will revert until fixed.

dshanske:
Still puzzled.

dshanske:
Here's the problem...

‘any‘ – retrieves any type except revisions and types with ‘exclude_from_search’ set to true.

function actionstream_init() {
	register_post_type('actionstream', array(
		'exclude_from_search' => true,
		'publicly_queryable' => true,
		'show_ui' => false,
		'label' => 'Actionstream',
		'taxonomies' => array('post_tag', 'category')
	));
}

dshanske:
https://core.trac.wordpress.org/ticket/17592 - Apparently it was a matter under discussion. Question is...how do we want to address it? For now, we could add a filter that allows custom post types to be expressly added, but it creates extra work. We could revert to directly querying the database...which means we lose the advantages of using WP_Query. 

I think maybe the best way is to call https://codex.wordpress.org/Function_Reference/get_post_types and pass in publicly_queryable = true instead of exclude_from_search = false and get our own list.

dshanske:
@singpolyma Can you change post_type to get_post_types( array( 'publicly_queryable' => true ) and see if that works for you?

singpolyma:
[...]

PR #94, opened 2016-08-06
Fix Issue 93

dshanske:
Related #93

### DIFF
--- a/includes/class-webmention-sender.php
+++ b/includes/class-webmention-sender.php
@@ -71,8 +71,7 @@ class Webmention_Sender {
 		);
 		if ( $webmention_server_url ) {
 			$response = wp_remote_post( $webmention_server_url, $args );
-		}
-		else {
+		} else {
 			$response = false;
 		}
 		// use the response to do something useful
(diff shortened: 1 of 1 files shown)


pfefferle:
this change has indention and syntax errors

dshanske:
I ran in through phpcbf. I thought it was supposed to catch that.

pfefferle:
this should be indented https://github.com/pfefferle/wordpress-webmention/pull/94/files#diff-5ac85cf40a7cd031c3978d311c357cdaR171

and there is a failing comma at the end https://github.com/pfefferle/wordpress-webmention/pull/94/files#diff-5ac85cf40a7cd031c3978d311c357cdaR175

dshanske:
I see the comma at the end, I didn't add that. Guess I need to hand check the output when I use auto correct.

dshanske:
That's it, I am taking @snarfed 's advice and trying to write unit tests.

pfefferle:
;)

this would be awesome!

dshanske:
I did some a while back based on endpoint discovery from webmentions.rock . I am going to try to write more in my 'playground'. I'd rather break things, perfect them, then try to move them to stable.

PR #104, opened 2016-10-29
Misc Document and Minor Tweaks to PR/100 Branch

dshanske:
Made some tweaks based on the refactoring. Brought back the async option as just a cron hook triggered by a flag.


pfefferle:
thanks a lot!

PR #115, opened 2017-01-27
 	frontpage processing for person tags to be accepted & setting default to sync

acegiak:


pfefferle:
this merge request is very confusing ;)

is the only change, the `sync` flag?

dshanske:
I think that @acegiak wants the Page ID for the homepage to be settable by user option, not just filter. I think that is doable.

acegiak:
Nono I ended up using the hook to do that in a different thread.

The only two things this PR are for are setting  the default to sync rather than async and allowing the webmention endpoint to be displayed on the front page so that you can receive person tags

dshanske:
@acegiak I see.. not showing it is creating an issue for exotic Webmentions.

dshanske:
I think I have another way to approach this.

dshanske:
Let me give it a go

pfefferle:
Is there perhaps a generic way? like an action or a hook? I don't want to add features that are not used by the core, but by a plugin instead.

dshanske:
That is what I am thinking. Make the is_singular and pings_open check a function, and filter it. So anyone can open it for any given URL.

PR #117, opened 2017-01-28
Improve Exotic Mentions Support

dshanske:
This makes it a bit easier for webmentions of archives, taxonomy, and home pages to be redirected to a specific post_id.

It moves 'webmention_post_id' to a global function so it can be retrieved outside of the receiver transaction. It adds a second function and filter, 'mentions_open' to determine if a URL is to be considered 'open' for these URLs to add the headers.

### DIFF
--- a/includes/class-webmention-receiver.php
+++ b/includes/class-webmention-receiver.php
@@ -128,9 +128,8 @@ class Webmention_Receiver {
 			return new WP_Error( 'target', __( 'Target is not on this domain', 'webmention' ), array( 'status' => 400 ) );
 		}
 
-		$comment_post_id = url_to_postid( $target );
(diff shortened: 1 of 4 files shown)


pfefferle:
I think this feature is worth the problems (https://wordpress.org/support/topic/endpoint-not-active/), so I removed it for the time being. Perhaps we re-think it and add something similar in the future.

pfefferle:
@dshanske anyway, thanks for your hard work!

dshanske:
@pfefferle Moving the filter outside of the receiver...regardless of removing the conditional check...seems worth it to improve it. Would you object to merging that portion of it?

pfefferle:
you mean the `webmention_post_id` filter?

pfefferle:
re-added! thanks for the hint!

PR #127, opened 2017-03-08
Minor Fixes

dshanske:
- Save Fragments of URLs
- Add Avatar Support to Webmentions
- Fix issue with dupes due incorrect query argument

### DIFF
--- a/includes/class-webmention-receiver.php
+++ b/includes/class-webmention-receiver.php
@@ -41,7 +41,7 @@ class Webmention_Receiver {
 	 * This is more to lay out the data structure than anything else.
 	 */
 	public static function register_meta() {
-		$args = array( 
+		$args = array(
 			'type' => 'string',
 			'description' => 'Target URL for the Webmention',
 			'single' => true,
@@ -59,8 +59,8 @@ class Webmention_Receiver {
 
 	/**
 	 * Show avatars on webmentions if set
- 	 *
- 	 * @param array $types list of avatar enabled comment types
+		 *
+		 * @param array $types list of avatar enabled comment types
 	 *
 	 * @return array show avatars also on trackbacks and pingbacks
 	 */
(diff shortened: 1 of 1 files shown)


pfefferle:
thanks a lot! I have only one point, I think we should use a more descriptive name for the fragment, like `webfinger_target_url_fragment` or `webfinger_url_fragment`... what do you think?

dshanske:
It isn't webfinger, it is webmention. But the longer name is fine.

dshanske:
I would say webmention_target_url and webmention_target_fragment

pfefferle:
I always mix these two names up... I like your suggestion!

pfefferle:
Can you add the changes or should I update the code after the merge?

dshanske:
Done.

pfefferle:
Sorry, but I think you broke something...

dshanske:
I will check. I did the code last night. I was exhausted. I thought I checked it this morning when I was awake, but must have missed something. Will double check.

dshanske:
I keep breaking my never commit while tired policy

pfefferle:
I found it... the slash in line 52 ;)

dshanske:
I don't see a slash in line 51 of the receiver? What do you see? I just tested it and it worked.

pfefferle:
sorry, my mobile app seems to show different line numbers. I commented inline.

dshanske:
I think I found the stray slash.

pfefferle:
sorry for beeing so pedantic, but can you also fix the codeclimate issues? https://codeclimate.com/github/pfefferle/wordpress-webmention/pull/127

Travis has problems with phpunit and php 5.6, so we don't have to care atm.

pfefferle:
Travis seems to be fixed now...

dshanske:
Fixed now, I think

pfefferle:
https://github.com/pfefferle/wordpress-webmention/pull/127/files#diff-4a405db849c60f60b9e32bde3bfe0096R62

dshanske:
I'm not sure what you are trying to convey.

pfefferle:
the indents are still wrong for that line and the following

PR #133, opened 2017-03-18
Webmention Supports

dshanske:
This PR adds a post type support feature called 'webmentions' which is automatically added to type 'post', added if a setting is set to 'page', and would require any custom post types to add or note during declaration.

This setting will open or close ping status by default. The sending of webmentions is now limited to post types that support this setting.

This allows a much clearer way to declare support for webmentions with custom post types and can be used to send/receive them. It also removes the issue created when WordPress disabled ping status support by default for pages and post types that didn't declare trackbacks support.

Because the logic of how support is maintained is now handled by the Webmentions plugin, if we decide to, in future, not rely on ping_status to determine webmention status...we can switch without requiring any changes outside the plugin.

### DIFF
--- a/templates/webmention-discussion-settings.php
+++ b/templates/webmention-discussion-settings.php
@@ -15,6 +15,14 @@
 
 	<br />
 
+	<label for="webmention_support_pages">
+		<input type="checkbox" name="webmention_support_pages" id="webmention_support_pages" value="1" <?php
+			echo checked( true, get_option( 'webmention_support_pages' ) );  ?> />
+		<?php _e( 'Enable Webmention Support for Pages', 'webmention' ) ?>
+	</label>
+
+	<br />
+
 	<label for="webmention_show_comment_form">
 		<input type="checkbox" name="webmention_show_comment_form" id="webmention_show_comment_form" value="1" <?php
 			echo checked( true, get_option( 'webmention_show_comment_form' ) );  ?> />
--- a/webmention.php
+++ b/webmention.php
@@ -33,6 +33,11 @@ class Webmention_Plugin {
 	 * Initialize Webmention Plugin
 	 */
 	public static function init() {
+		// Add a new feature type to posts for webmentions
+		add_post_type_support( 'post', 'webmentions' );
+		//	if ( 1 == get_option( 'webmention_support_pages' ) ) {
+			add_post_type_support( 'page', array( 'webmentions' ) );
+		//	}
 		if ( WP_DEBUG ) {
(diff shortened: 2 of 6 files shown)


webmention.php
```diff
@@ -74,6 +94,12 @@ public static function admin_register_settings() {
 			'show_in_rest' => true,
 			'default' => 0,
 		) );
+		register_setting( 'discussion', 'webmention_support_pages', array(
```
pfefferle: this is not used, so we shouldn't add it

webmention.php
```diff
@@ -33,6 +33,11 @@ class Webmention_Plugin {
 	 * Initialize Webmention Plugin
 	 */
 	public static function init() {
+		// Add a new feature type to posts for webmentions
+		add_post_type_support( 'post', 'webmentions' );
+		//	if ( 1 == get_option( 'webmention_support_pages' ) ) {
```
pfefferle: please don't commit debug or unused stuff

templates/webmention-discussion-settings.php
```diff
@@ -15,6 +15,14 @@
 
 	<br />
 
+	<label for="webmention_support_pages">
```
pfefferle: this is not used, so we shouldn't add it

pfefferle:
plugin requires WP 4.7 so we should update the travis config to not test versions below 4.7

webmention.php
```diff
@@ -33,6 +33,11 @@ class Webmention_Plugin {
...
+			add_post_type_support( 'page', array( 'webmentions' ) );
```
pfefferle: why is `webmentions` for posts a string and for pages an array?

webmention.php
dshanske: That is my fault. I forgot to uncomment before I commented. I was debugging and forgot to remove the comment I put in while I was testing the feature.

webmention.php
dshanske: No reason. Either a string or an array is acceptable. Will simplfy.

dshanske:
I took out the comments I forgot to before committing and it works as advertised. Can you rereview?

pfefferle:
thanks :)

pfefferle:
nice idea btw!

dshanske:
Well, when I was trying to solve the custom post type issue @techlifeweb raised, I realized we missed that the turning off of pings by default in 4.3 had consequences that affected pages and custom post types. This seemed like the best and most flexible way around it.

PR #136, opened 2017-03-27
Support Creation Time and Webmention Source Storage

dshanske:
This is to support storing webmention source and creation time in meta. 

The reasoning is that creation time is overwritten by Semantic Linkbacks and allows for the original time to be compared to the modified time, which is not stored in the comment table.

The source is stored, like pingbacks, in the author_url, which results in duplication of purpose, especially in Semantic Linkbacks. The code will still look at the comment_author_url if there is no matching meta property. 

Realized during this that the fragment and target code previously written executes after Semantic Linkbacks, so had to move everything earlier in the chain of events. 

The goal is to have webmentions move a bit away from the pingback mirrored implementation, but still keeping backward compatibility with that. That said, there is a proposal(https://core.trac.wordpress.org/ticket/17771) that discusses this issue for a different reason...

### DIFF
--- a/includes/class-webmention-receiver.php
+++ b/includes/class-webmention-receiver.php
@@ -29,12 +29,6 @@ class Webmention_Receiver {
 		add_filter( 'webmention_comment_data', array( 'Webmention_Receiver', 'default_title_filter' ), 21, 1 );
 		add_filter( 'webmention_comment_data', array( 'Webmention_Receiver', 'default_content_filter' ), 22, 1 );
 
-		// Save Fragments
-		add_filter( 'preprocess_comment', array( 'Webmention_Receiver', 'save_fragment' ), 9, 1 );
-
-		// Save Original Creation Time in GMT
-		add_filter( 'preprocess_comment', array( 'Webmention_Receiver', 'save_original_time' ), 9, 1 );
-
 		// Allow for avatars on webmention comment types
 		add_filter( 'get_avatar_comment_types', array( 'Webmention_Receiver', 'get_avatar_comment_types' ) );
 		self::register_meta();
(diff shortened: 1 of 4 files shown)


pfefferle:
we don't remove the fragment from the url do we? if we don't, we shouldn't add the fragment query, because it is part of the url and it would break backwards compatibility.

dshanske:
The search is for the source URL and the post ID. Without adding in the fragment, it might update two fragments to the same URL

pfefferle:
but I thought we safe the source with the fragment... don't we? and if we do, we already support fragments without extracting it...

includes/class-webmention-receiver.php
```diff
@@ -382,25 +407,57 @@ public static function check_dupes( $commentdata ) {
...
+			$commentdata['comment_approved'] = $comment->comment_approved;
 
+			return $commentdata;
+		}
+
+		// Check in comment_author_url if the newer location is empty
 		$args = array(
 			'post_id' => $commentdata['comment_post_ID'],
-			'author_url' => esc_url_raw( $commentdata['comment_author_url'] ),
+			'author_url' => $commentdata['comment_author_url'],
```
pfefferle: can we remove the filter, because we use it for saving the data!?!

dshanske:
The saving data information is now built in. But this is the target fragment we're checking.

includes/class-webmention-receiver.php
dshanske: If you are referring to esc_url_raw...we actually call it when we set comment_author_url.

pfefferle:
ok, now I got it!

dshanske:
Does it make sense?

dshanske:
Thank you, @pfefferle. You made my day.

PR #153, opened 2017-06-30
Some Fixes

dshanske:
Adds a webmention meta box to the comment edit screen so these fields can be viewed, adds a filter to the webmention_comment_form, adds an edit_webmention hook due to the fact the edit_comment filter doesn't allow for meta updates at this time(needed for Semantic Linkbacks).


pfefferle:
thanks

PR #159, opened 2017-11-18
Phpcs

dshanske:
This adds the PHPCS and Travis Changes I implemeneted in the WordPress Indieweb plugin. That includes testing for version 5.2 and over, textdomain checking, etc.


pfefferle:
Thanks! 😊

PR #160, opened 2017-12-08
Add localization and German language files

deponeWD:
To use the plugin seamlessly on sites directed to German speaking audiences I have added localization and created a first translation.


pfefferle:
Oh, thanks!

PR #164, opened 2018-01-25
French translation

edas:
I am not used to software translation and I would welcome any review before considering this a quality work, but it is probably better than nothing (at least, public pages in French won't show a webmention form with English text)


pfefferle:
Mercie 😍

PR #169, opened 2018-05-07
Form Improvements

dshanske:
Made some fixes.

### DIFF
--- a/includes/class-webmention-admin.php
+++ b/includes/class-webmention-admin.php
@@ -53,10 +53,26 @@ class Webmention_Admin {
 	 * @param int $comment_id the comment id
 	 */
 	public static function manage_comments_custom_column( $column, $comment_id ) {
-		$comment_type = get_comment_type( $comment_id );
-		if ( 'comment_type' === $column ) {
-			_e( $comment_type, 'webmention' );
+		if ( 'comment_type' !== $column ) {
+			return;
 		}
+		$type = get_comment_type( $comment_id );
+		switch ( $type ) {
+			case 'trackback':
+				_e( 'Trackback', 'webmention' );
+				break;
+			case 'pingback':
+				_e( 'Pingback', 'webmention' );
+				break;
+			case 'comment':
+				_ex( 'Comment', 'noun', 'webmention' );
+				break;
+			case 'webmention':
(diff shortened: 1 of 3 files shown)


includes/functions.php
```diff
@@ -18,7 +18,7 @@ function send_webmention( $source, $target ) {
  *
  */
 function get_webmention_form_text( $post_id ) {
-	return apply_filters( 'webmention_form_text', __( 'Respond on your own site? Send me a <a href="http://indieweb.org/webmention">Webmention</a> by writing something on your website that links to this post and then enter your post URL below.', 'webmention' ), $post_id );
+	return apply_filters( 'webmention_form_text', __( 'Respond on your own site, link to this post and enter the URL below to link it below. Want to update or remove the mention? Just enter it again to update.(<a href="http://indieweb.org/webmention">Learn More</a>)', 'webmention' ), $post_id );
```
pfefferle: Can you try to rephrase that? I wouldn't use `below` two times here: `enter the URL below to link it below`

includes/class-webmention-admin.php
```diff
@@ -155,13 +161,26 @@ public static function comment_columns( $columns ) {
...
-		add_options_page(
-			'Webmention',
-			'Webmention',
-			'manage_options',
-			'webmention',
-			array( 'Webmention_Admin', 'settings_page' )
-		);
+		$title = __( 'Webmention', 'webmention' );
+		// If the IndieWeb Plugin is installed use its menu.
+		if ( class_exists( 'IndieWeb_Plugin' ) ) {
```
pfefferle: thinking of usability... shouldn't we always add both? I created a settings page, because this was feedbacked a lot of times... I am not sure if everyone searches in the IndieWeb area!?! what do you think?

perhaps @chrisaldrich can help here?

includes/functions.php
dshanske: I wouldn't mind some brainstorming on this, maybe some input from @chrisaldrich and others. But the original focused on the plumbing(webmention), over the user benefit. And it needs to tell people they can delete or update with the form to cover some of the privacy implications.

dshanske: Tried another phrasing. Thoughts?

Issue #166, opened 2018-04-06
Prevent webmentions sending for specified links


colin-walker:
The webmention plugin sends mention for all links in a post but there may be times when you don’t want a webmention sent, especially when dealing with social properties like micro.blog or Mastodon where a mention may create duplication if your post is going to be POSSE’d over.

How about a mechanism to specify on a link that you don’t want it to trigger a mention?

A site blacklist is too extreme as this precludes ever sending to that domain so here’s a suggestion:

We could add `class=“nomention”` to the link then check for this in the plugin using the following:

Replace `$links = wp_extract_urls( $post->post_content );` within `public static function send_webmentions` with

```
preg_match_all('/<a[^>]+>/i',$post->post_content, $results); 

$mentions = '';
foreach ($results[0] as $link) {
  if (!strpos($link, 'nomention')) {
    $mentions .= $link;
  }
}

$links = wp_extract_urls($mentions);
```

to cycle through the <a> tags in the post, add those _without_ the nomention class to a new string and extract the URLs from that strong instead.

dshanske:
Isn't this basically rel=nofollow ?

pfefferle:
This limits the parser to links and ignores images, videos, ... It is also not very user friendly, because nearly noone knows about microformats and writes plain HTML. Besides that, I do not think this is really an issue that needs to be solved on the WordPress side, only if micro.blog does not handle the Webmentions well.

colin-walker:
Yes, I was thinking about it like nofollow but that doesn’t impact webmentions does it?

Good point, about images & videos - that could be extended.

It was just a quick thought about generating a mechanism to do it.

pfefferle:
For my better understanding, is this really an issue or is it only because of the handling of micro.blog?

colin-walker:
The micro.blog incident was what triggered the thinking but surely they may be other times when you don’t necessarily want/need a mention generated. Just thinking out loud.

pfefferle:
I would prefer a more [gen 2 - 4](https://indieweb.org/generations) friendly solution, if this is really a problem that needs to be solved.

pfefferle:
@colin-walker but is this really link specific, or is a handling on blog level also a solution?

colin-walker:
As I say, it was just an initial idea based on the plugin itself. If there was a wider need then, yes, there would no doubt be a better way of implementing it at the spec level.

pfefferle:
[...]

PR #170, opened 2018-05-08
Domain Whitelist

dshanske:


includes/class-webmention-admin.php
```diff
@@ -297,6 +297,14 @@ public static function register_settings() {
 				'default'      => 0,
 			)
 		);
+		register_setting(
+			'webmention', 'whitelist_domains', array(
+				'type'         => 'array',
+				'description'  => __( 'Domain Whitelist for Linkbacks', 'webmention' ),
```
pfefferle: we shouldn't use `Linkback` in this case, because it is only for Webmentions

includes/class-webmention-receiver.php
```diff
@@ -711,24 +715,66 @@ public static function publish_post_hook( $post_id ) {
...
-		$commentdata = self::check_dupes($commentdata);
+		$commentdata = self::check_dupes( $commentdata );
 
 		if ( isset( $commentdata['comment_ID'] ) ) {
 			wp_delete_comment( $commentdata['comment_ID'] );
 		}
 	}
 
+
+	public static function whitelist_approved( $commentdata ) {
```
pfefferle: we discussed a lot of whitelists/blacklists lately (for example a blacklist, to disable sending webmentions to a specific domain https://github.com/pfefferle/wordpress-webmention/issues/166), so we should make the functions more descriptive. At least we should add something like `receiver` to the name.

includes/class-webmention-receiver.php
```diff
@@ -711,24 +715,66 @@ public static function publish_post_hook( $post_id ) {
...
+	public static function whitelist_check( $url ) {
+		$whitelist = get_option( 'whitelist_domains' );
+		$whitelist = trim( $whitelist );
+		$host      = wp_parse_url( $url, PHP_URL_HOST );
+		// strip leading www, if any
+		$host = preg_replace( '/^www\./', '', $host );
+		if ( '' === $whitelist ) {
+			return false;
+		}
+		$domains = explode( ',', $whitelist );
```
pfefferle: the blacklist/whitelists of WordPress are using "one item per line" instead of comma seperated, so we should be consistent.

templates/webmention-settings.php
```diff
@@ -53,6 +53,14 @@
 				}
 				?>
 			</label>
+
+			<br />
+
+			<label for="whitelist_domains">
```
pfefferle: We should add a description what this feature is doing, like

> A Webmention from a page, that matches a domain in this list will be auto-approved. One domain (e.g. `indieweb.org`) per line.

dshanske:
All proposed changes make sense. Will implement and update tonight.

dshanske:
This should address the issue

PR #173, opened 2018-05-14
changed text

pfefferle:


dshanske:
The profile picture is a Semantic Linkbacks feature.

pfefferle:
And the rest?

dshanske:
It's accurate

PR #180, opened 2018-06-02
Settings

dshanske:
This splits the settings page into pieces so the Semantic Linkbacks piece can be easily inserted inline.


pfefferle:
I don't think this change is needed, because all the template stuff seems only be added because of the extensibility with `do_action` but settings can be extended with `do_settings_fields( 'webmention', 'sender' );`, `do_settings_fields( 'webmention', 'receiver' );` and `do_settings_sections( 'webmention' );`.

PR #181, opened 2018-07-01
Receiving Vouches

dshanske:
This PR adds in the ability to add domains to the whitelist from the comment admin,  as well as preliminary support for receiving vouches.

For now, the code, if there is a vouch parameter, verifies it. If it fails verification, it rejects the webmention. If it passes, it adds a parameter noting it passed. It also stores the vouch. It uses the approved domains list to determine acceptable sites for vouching, but I'd like to expand verification in future.

The next piece would be to add an admin setting to decide whether or not a successful vouch should be auto-approved, and to start getting some vouched webmentions.

### DIFF
--- a/includes/class-webmention-receiver.php
+++ b/includes/class-webmention-receiver.php
@@ -26,11 +26,14 @@ class Webmention_Receiver {
 		add_filter( 'webfinger_post_data', array( 'Webmention_Receiver', 'jrd_links' ) );
 
 		// Webmention helper
+		add_filter( 'webmention_comment_data', array( 'Webmention_Receiver', 'webmention_vouch' ), 10, 1 );
 		add_filter( 'webmention_comment_data', array( 'Webmention_Receiver', 'webmention_verify' ), 11, 1 );
 		add_filter( 'webmention_comment_data', array( 'Webmention_Receiver', 'check_dupes' ), 12, 1 );
 
 		// Webmention whitelist
 		add_filter( 'webmention_comment_data', array( 'Webmention_Receiver', 'auto_approve' ), 13, 1 );
+		add_filter( 'comment_row_actions', array( 'Webmention_Receiver', 'comment_row_actions' ), 13, 2 );
+		add_filter( 'comment_unapproved_to_approved', array( 'Webmention_Receiver', 'transition_to_whitelist' ), 10 );
 
 		// Webmention data handler
 		add_filter( 'webmention_comment_data', array( 'Webmention_Receiver', 'default_title_filter' ), 21, 1 );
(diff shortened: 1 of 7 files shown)


pfefferle:
Vouch is very experimental and (as you said) there are only few who send vouches, so I am not sure if we sould add this in such an early state. This might cause errors on verifications and might result in timeouts, because of the extra site that has to be parsed. I would prefer implementing async Webmentions first and/or implementing Vouch as a feature plugin or behind a feature toggle (so users can enable it using for example a constant).

pfefferle:
Cool. I think the best way to do so, is to encapsulate all in a `class-webmention-vouch.php` file. With all the hooks and settings (if needed). So that we simply can enable/disable vouch via:

```php
if ( WEBMENTION_VOUCH === true ) {
    require_once dirname( __FILE__ ) . '/includes/class-webmention-vouch.php';
}
```

What do you think?

pfefferle:
nice, thanks! can you rename the `vouch` method to a more descriptive one? For example `verify` or `verify_vouch`. So that we can differ beteen sending and verifying...

dshanske:
@pfefferle Done if you want to review

includes/class-webmention-receiver.php
```diff
@@ -96,6 +133,15 @@ public static function register_meta() {
 			'show_in_rest' => true,
 		);
 		register_meta( 'comment', 'webmention_response_code', $args );
+
+		// Purpose of this is to store a vouch URL
```
pfefferle: Can you move this to the vouch class, otherwise it will load, if vouch is enabled or not...

includes/class-webmention-receiver.php
```diff
@@ -50,6 +53,40 @@ public static function init() {
...
+	public static function transition_to_whitelist( $comment ) {
+		if ( ! current_user_can( 'moderate_comments' ) ) {
+			return;
+		}
+		if ( isset( $_REQUEST['domain'] ) ) {
+			$url = get_comment_meta( $comment->comment_ID, 'webmention_source_url', true );
+			if ( ! $url ) {
+				return;
+			}
+			$whitelist = get_option( 'webmention_approve_domains' );
```
pfefferle: can you encapsulate this in some functions? For example `get_webmention_approve_domains` and `add_webmention_approve_domains`... (just example names, feel free to change them to a better/more readable version)

includes/class-webmention-receiver.php
```diff
@@ -50,6 +53,40 @@ public static function init() {
...
+			$whitelist = trim( $whitelist );
+			$whitelist = explode( "\n", $whitelist );
+			$host      = wp_parse_url( $url, PHP_URL_HOST );
+
+			// strip leading www, if any
```
pfefferle: see comment above

includes/class-webmention-receiver.php
```diff
@@ -387,6 +438,39 @@ public static function post( $request ) {
 		return new WP_REST_Response( $return, 200 );
 	}
 
+	/**
```
pfefferle: should be also part of the vouch class?

includes/class-webmention-receiver.php
dshanske: I'd like to store the vouch URL whether or not vouch is enabled or not, in the event it becomes stable in future.

includes/class-webmention-receiver.php
dshanske: Will do.

includes/class-webmention-receiver.php
dshanske: Back to the question I asked re storage.

pfefferle: OK, will think about that, but then I have one more question: Why not directly set the comment-meta in line 318 and following and move it above the commentdata setter in line 316?

dshanske: Because there might not be a vouch parameter? It is optional.

pfefferle: But this is what you also check in line 318...

pfefferle: why not:

```php
if ( isset( $params['vouch'] ) ) {
	$comment_meta['vouch'] = esc_url( urldecode( $params['vouch'] ) );
}
```

dshanske: I misunderstood...you are right.

pfefferle: like you do with the date: https://github.com/pfefferle/wordpress-webmention/pull/181/files#diff-4a405db849c60f60b9e32bde3bfe0096R308

dshanske:
@pfefferle Requesting a re-review

dshanske:
@pfefferle Can you merge and do a release then?

PR #205, opened 2019-08-02
Julfixes

dshanske:
This enables webmentions by default on the theory if you enable the plugin, you must want them, so it shouldn't be triggered by the pingback flag setting.

Also fixes the auto approve whitelist and a few other minor tweaks.

### DIFF
--- a/includes/class-webmention-admin.php
+++ b/includes/class-webmention-admin.php
@@ -23,8 +23,6 @@ class Webmention_Admin {
 		add_filter( 'manage_edit-comments_columns', array( 'Webmention_Admin', 'comment_columns' ) );
 		add_filter( 'manage_comments_custom_column', array( 'Webmention_Admin', 'manage_comments_custom_column' ), 10, 2 );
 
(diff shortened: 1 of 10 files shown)


pfefferle:
@dshanske it seems that you have made your changes to an old revision and that caused some conflicts.

includes/class-webmention-admin.php
```diff
@@ -23,8 +23,6 @@ public static function init() {
 		add_filter( 'manage_edit-comments_columns', array( 'Webmention_Admin', 'comment_columns' ) );
 		add_filter( 'manage_comments_custom_column', array( 'Webmention_Admin', 'manage_comments_custom_column' ), 10, 2 );
 
-		// Webmention whitelist
```
pfefferle: oops... thanks for fixing my c&p issue!

dshanske:
I will see about rebasing

dshanske:
Should be fixed now. Sorry.

pfefferle:
Thanks :)

PR #212, opened 2019-08-08
Force setting headers

Lewiscowles1986:
Hacking through this in a browser was not fun http://sandbox.onlinephpfunctions.com/code/3c60598139a7c059643400b9bb0a6dd2ce965f1a shows the two cases I REPL'd

being in a class may lead to iteration.

* uses force argument (should clobber)
* gets existing header values (should make clobbering not a problem)
* has to filter header values due to PHP API
* sets single header supporting multiple values
* adds guard clause in-case headers are sent

### DIFF
--- a/includes/class-webmention-receiver.php
+++ b/includes/class-webmention-receiver.php
@@ -837,9 +837,44 @@ class Webmention_Receiver {
 	 * The Webmention autodicovery http-header
 	 */
 	public static function http_header() {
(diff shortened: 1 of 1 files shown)


Lewiscowles1986:
Do not merge. This is not needed. Had I performed on a greater sample size than one post, one time in 3 clients I might have noticed that the headers are there.

PR #213, opened 2019-08-09
Avatar

dshanske:
Add avatar handling support into Webmentions. This can coexist with the handler in Semantic Linkbacks and take over if that is disabled. If the Semantic Linkbacks one is working, this one will never be triggered.


dshanske:
@pfefferle What do you think?

pfefferle:
I think `Avatar_Handler` might be a bit too generic...

pfefferle:
But I will have a look later today

dshanske:
I am not against changing the name. My goal is to move Semantic Linkbacks functionality piece by piece. So this means that the icons won't vanish if you disable the other plugin.

dshanske:
@pfefferle What name would you suggest?

pfefferle:
Just prefixing it with „Webmention“ to have no naming collisions.

dshanske:
@pfefferle Prefixed now.

dshanske:
@pfefferle Any thought on this?

includes/class-webmention-avatar-handler.php
```diff
@@ -0,0 +1,170 @@
...
+			$args['url'] = self::get_default_avatar( $args['default'] );
+			return $args;
+		}
+		if ( ! strpos( $args['url'], 'gravatar.com' ) ) {
+			return $args;
+		}
+		$args['url'] = str_replace( 'd=' . $args['default'], 'd=' . self::get_default_avatar( $args['default'] ), $args['url'] );
+		if ( ! empty( $comment->comment_author_email ) ) {
+			if ( self::check_gravatar( $comment ) ) {
+				update_comment_meta( $comment->comment_ID, 'avatar', $args['url'] );
```
pfefferle: does this store the local anonymous avatar to the db? I would avoid storing anonymous or temporary data to the db, because that means a lot of migration when we have/want to change the avatar and it makes it impossible to use a different one.

dshanske: Yes, it's how Semantic Linkbacks is doing it now.. but will think of another solution based on that

includes/class-webmention-avatar-handler.php
```diff
@@ -0,0 +1,170 @@
...
+	 */
+	public static function get_avatar_meta( $comment ) {
+		if ( is_numeric( $comment ) ) {
+			$comment = get_comment( $comment );
+		}
+		if ( ! $comment ) {
+			return false;
+		}
+		$avatar = get_comment_meta( $comment->comment_ID, 'avatar', true );
+		// Backward Compatibility for Semantic Linkbacks
```
pfefferle: nice!

dshanske:
@pfefferle Substituted saving the default avatar in meta with caching the lack of a gravatar as a transient for 24 hours, and only if there is an email. Otherwise it doesn't look for one.

dshanske:
Wondering if 24 hours is too long.

dshanske:
On the other hand, it is only caching whether or not they have a gravatar, not what it is

dshanske:
@pfefferle Made the caching time configurable here now.

dshanske:
@pfefferle Anything else to change here?

dshanske:
Release time?

PR #218, opened 2019-08-30
lets escape as much as we can

tw2113:
There are some outliers of strings that have links in the middle of paragraphs, but this should cover majority of the non-escaped translated strings.


pfefferle:
I really like it, but it is hard to review on the phone. @dshanske can you help?

pfefferle:
Nice, thanks!

tw2113:
even though we hope the errors don't show up...ever, we translate them for a reason, so may as well safely translate too.

PR #221, opened 2019-09-04
Attachments

dshanske:
This started as simply adding attachment support, but in order to support that, I upgraded the setting to allow one to add webmention support to any public post type.

webmention_to_post_id now tries to match attachment files to the attachment post id. 

In exploring this, and realizing that I couldn't support receiving webmentions to media files without adding the header on the server level(Thanks @pfefferle for leading me to that conclusion when it eluded me), I looked at the fact that we were sending out webmentions for every link in a post, and links in img, video, or audio tags would almost never be relevant, and there were other HTML elements that it made no sense to send webmentions for.

So now, it actually extracts valid links, not just URLs in content.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,12 +31,6 @@ matrix:
     env: WP_VERSION=4.7 WP_MULTISITE=0
   - php: 5.4
     env: WP_VERSION=4.7 WP_MULTISITE=0
-  - php: 5.3
-    env: WP_VERSION=4.7 WP_MULTISITE=0
-    dist: precise
-  - php: 5.2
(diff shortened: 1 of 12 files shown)


dshanske:
Oh, yes...also two other notes. Bumped minimum PHP to 5.4, which is still well below current version of 5.6, but is minimum officially supported for PHP-MF2 and namespaces.

While I did add a notation if the HTML5 DOMDocument replacement is there(same coding that appears in PHP-MF2), that it would use it, but have no intention of bundling it.

pfefferle:
thanks a lot @dshanske

Issue #152, opened 2017-06-30
Endpoint advertised on homepage even if not accepted


colin-walker:
Now that options exist to enable/disable webmentions on the homepage should the endpoint no longer be advertised on the home page/other pages that don't accept webmentions?

dshanske:
We tried that. There were issues and it was reverted.

colin-walker:
Fair enough, it was just a thought.

dshanske:
I am not closing it as an issue. Just mentioning the challenge

dshanske:
I can't remember what the problem was. Thinking of bringing back this feature and it could be disabled.

dshanske:
Related #75

dshanske:
Addressed now

Issue #75, opened 2016-04-13
Remove webmention header if Pings closed


dshanske:
If a page doesn't support pings/webmentions then the endpoint shouldn't be advertised on that page.

dshanske:
https://github.com/Automattic/_s/pull/965

Further to this idea...it looks like themes are doing something similar for pings.

pfefferle:
good point

pfefferle:
done

snarfed:
sounds like this was done but then later reverted? i'd like this. my use case is that i'm using "unlisted" posts to approximate private posts with better accessibility. background: https://indieweb.org/unlisted#Similarity_to_private_posts

i don't want to encourage or accept webmentions on these posts, so i'd like to turn off the webmention endpoint `<link>` tag and HTTP header. the posts are all in a specific category, so triggering on that would be an alternative, but i expect i can just disable pings on all posts in that category.

@dshanske shall we reopen this?

dshanske:
Related #152

dshanske:
I still can't remember what the problems were. @pfefferle Do you remember exactly what the issue was so I can try to address in a new option?

dshanske:
We keep revisiting this conversation, but I keep forgetting to document it

dshanske:
Merged in again. New implementation

PR #223, opened 2019-09-06
Hide headers if webmentions are not supported to that URL

dshanske:
Address #75 and #152 

The problem with hiding headers prior was partially that is_singular() or anything requiring a global query variable isn't loaded when headers are sent, so it would always evaluate to false.

This uses get_self_link(scheduled for WordPress 5.3, see https://core.trac.wordpress.org/ticket/44838) to retrieve the URL currently being requested, then determines the post_id if applicable using webmention_to_postid, which means the existing filter could take archive or other pages and redirect them. It also works for the currently built in home page mention redirect.

### DIFF
--- a/includes/class-webmention-receiver.php
+++ b/includes/class-webmention-receiver.php
@@ -829,28 +829,36 @@ class Webmention_Receiver {
 	 *
 	 * @return boolean
 	 */
-	public static function receive_mentions() {
-		if ( 0 !== WEBMENTION_ALWAYS_SHOW_HEADERS ) {
+	public static function should_show_headers() {
+		if ( WEBMENTION_ALWAYS_SHOW_HEADERS ) {
 			return true;
 		}
+
+		return self::should_receive_mentions();
+	}
+
+	/**
+	 *
+	 *
+	 * @return boolean
+	 */
+	public static function should_receive_mentions() {
 		if ( ! is_singular() ) {
-			// Cache this as will be called twice in a page load
-			$post_id = wp_cache_get( base64_encode( get_self_link() ), 'wmurl' );
-			if ( false === $post_id ) {
-				$post_id = webmention_url_to_postid( get_self_link() );
-				wp_cache_set( base64_encode( get_self_link() ), $post_id, 'wmurl', 300 );
-			}
+			$post_id = webmention_url_to_postid( get_self_link() );
+
 			if ( ! $post_id ) {
 				return false;
 			}
 		}
 		// If the post type does not support webmentions do not even check if pings_open is set
 		if ( ! post_type_supports( get_post_type( $post_id ), 'webmentions' ) ) {
-				return false;
+			return false;
 		}
+
 		if ( pings_open( $post_id ) ) {
 			return true;
 		}
+
 		return false;
 	}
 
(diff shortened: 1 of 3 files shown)


pfefferle:
What about homepage webmentions?

pfefferle:
In the WebSub plugin I use the `template_redirect` hook, which supports the `is_*` functions https://github.com/pubsubhubbub/wordpress-pubsubhubbub/blob/master/includes/class-pubsubhubbub-topics.php#L48

dshanske:
homepage mentions are covered because if not set, webmention url to post id will return 0

pfefferle:
But then it returns false and adds no header!?!

pfefferle:
What I mean is, it shows no headers on the frontage!

pfefferle:
Have you had a look at the websub code?

pfefferle:
And what about static front pages?

pfefferle:
And what about custom pages/archive pages https://github.com/pfefferle/wordpress-webmention#how-can-i-handle-webmentions-to-my-homepage-or-archive-pages

pfefferle:
This wouldn‘t work any more...

dshanske:
It would. I checked it. Because it would rely on the page it was redirecting to having pings open

dshanske:
So, the filter redirects archive/home pages to a post id. The function checks that post id for whether it accepts pings. If there is no post id, it. would return false

pfefferle:
Thanks for the clarification on the #indieweb chat! Looks promising!

pfefferle:
Perhaps we can move the pings_open check to the `webmention_url_to_postid` function?

dshanske:
That would retrieve the entire post every time we retrieved the page... although that technically would happen anyway.

dshanske:
I moved it, as suggested to the template_redirect hook as otherwise it would generate a WP_Query to find the post on every page load, and that could get expensive. Also added caching for when it is an archive page so that it would only run the query once per page load, as opposed to twice.

dshanske:
Also added the constant to disable it.

includes/class-webmention-receiver.php
```diff
@@ -824,10 +824,43 @@ public static function comment_template_include( $template ) {
...
+	 *
+	 * @return boolean
+	 */
+	public static function receive_mentions() {
+		if ( 0 !== WEBMENTION_ALWAYS_SHOW_HEADERS ) {
+			return true;
+		}
+		if ( ! is_singular() ) {
+			// Cache this as will be called twice in a page load
+			$post_id = wp_cache_get( base64_encode( get_self_link() ), 'wmurl' );
```
pfefferle: But this will only work if you have installed a cache plugin: https://codex.wordpress.org/Class_Reference/WP_Object_Cache#Persistent_Caching

dshanske: Must disagree. That is if you want it to persist for multiple page loads. I want it to persist during the same page load...once for the http headers and once for the html headers

pfefferle: Ah I see: https://codex.wordpress.org/Class_Reference/WP_Object_Cache#Examples

dshanske: If someone enables a persistent cache, then they'd derive an extra benefit

pfefferle: I learned a lot today! thanks!

pfefferle:
thanks a lot @dshanske

PR #224, opened 2019-09-07
Improve Storage of Parameters

dshanske:
The primary  for this PR was to address the compact issue, where newer versions of PHP throw a warning if the variable you are running compact with isn't declared.

It introduces a new function, hooked onto the webmention_comment_data filter, that takes parameters passed in commentdata and stores them in the comment_meta array for saving into the database. It means it could be removed or replaced, or alternatively, the mapping list could be filtered in future.

As part of improving comment meta, I added storage of the response code, which had a registered meta key but wasn't being stored, moved the created key, which has yet to be used, which stored original webmention creation as opposed to what was later parsed, and started proactively adding the protocol key for webmention, which matches the activitypub plugin. 

I updated the meta box in the edit comment screen and noticed that we had two versions of it, so removed one.

### DIFF
--- a/includes/class-webmention-receiver.php
+++ b/includes/class-webmention-receiver.php
@@ -35,6 +35,9 @@ class Webmention_Receiver {
 
 		add_filter( 'webmention_comment_data', array( 'Webmention_Receiver', 'auto_approve' ), 13, 1 );
 
+		// Store extra parameters in metadata
+		add_filter( 'webmention_comment_data', array( 'Webmention_Receiver', 'store_meta' ), 99, 1 );
+
 		// Allow for avatars on webmention comment types
 		if ( 0 !== (int) get_option( 'webmention_avatars', 1 ) ) {
 			add_filter( 'get_avatar_comment_types', array( 'Webmention_Receiver', 'get_avatar_comment_types' ), 99 );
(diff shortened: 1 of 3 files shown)


includes/class-webmention-receiver.php
```diff
@@ -264,47 +311,42 @@ public static function post( $request ) {
...
+		$protocol          = 'webmention';
 
 		// change this if your theme can't handle the Webmentions comment type
 		$comment_type = WEBMENTION_COMMENT_TYPE;
 
 		// change this if you want to auto approve your Webmentions
 		$comment_approved = WEBMENTION_COMMENT_APPROVE;
 
-		$commentdata = compact( 'comment_type', 'comment_approved', 'comment_agent', 'comment_date', 'comment_date_gmt', 'comment_meta', 'source', 'target', 'vouch' );
+		$commentdata = compact( 'protocol', 'comment_type', 'comment_approved', 'comment_agent', 'comment_date', 'comment_date_gmt', 'created', 'source', 'target', 'vouch' );
```
pfefferle: do we still need this? I thought this was only to prefill the array, but with https://github.com/pfefferle/wordpress-webmention/pull/224/files#diff-4a405db849c60f60b9e32bde3bfe0096R232 shouldn't it be obsolete?

dshanske: Actually no. This still prefills the array. The store stores extra parameters in meta...the ones not stored in the comment table. Previously, our extra parameters were stored at the beginning rather than the end.

dshanske: Any other questions?

pfefferle: But wasn't your initial problem, that `compact` does not work in newer PHP versions?

dshanske: There were two. The other was allowing more flexibility in the storage of parameters.

dshanske: The compact pre-fills the array with properties. The storing of them in meta, the ones that aren't part of the comment table, is now done in the new function.

includes/class-webmention-receiver.php
```diff
@@ -210,6 +222,41 @@ public static function get( $request ) {
...
 	}
 
+	/**
+	 * Store extra parameters in comment meta
+	 *
+	 * @param array $commentdata
+	 *
+	 * @return array|null|WP_Error $commentdata The Comment Array with added comment meta field or a WP_Error object.
+	 */
+	public static function store_meta( $commentdata ) {
```
pfefferle: why not directly set them to `$commentdata['comment_meta']`? would save us a lot of code.

dshanske: I wanted to separate out the comment meta storage from the comment data storage. Mixing them was somewhat confusing.

dshanske: I can try to reorder it differently.

dshanske: I am going to split off the non controversial parts of this and take a separate shot at the other piece

PR #225, opened 2019-09-16
Verify Using Media Type Rules

dshanske:
https://www.w3.org/TR/webmention/#webmention-verification

`The receiver SHOULD use per-media-type rules to determine whether the source document mentions the target URL. For example, in an [ HTML5] document, the receiver should look for <a href="*">, , <video src="*"> and other similar links.`

This also passes the DOMDocument object into the commentdata array so that PHP-MF2 can use it in a future iteration.

It includes @aaronpk's build_url and a normalizing function for URLs to help in comparisons.

Also, discovered another issue causing a failure, an inverted conditional in webmention_to_urlid.

### DIFF
--- a/includes/class-webmention-receiver.php
+++ b/includes/class-webmention-receiver.php
@@ -230,13 +230,13 @@ class Webmention_Receiver {
 			return new WP_Error( 'source_missing', esc_html__( 'Source is missing', 'webmention' ), array( 'status' => 400 ) );
 		}
 
-		$source = urldecode( $params['source'] );
+		$source = self::normalize_url( urldecode( $params['source'] ) );
 
 		if ( ! isset( $params['target'] ) ) {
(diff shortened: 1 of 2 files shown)


pfefferle:
Sorry, but I do not understand the advantage of this change. In the end it seems to make it way more complex and adds no advantages, because it does not really treat things differently. The check at this time of the process is only to check if there is a link, the real check, should be later at the real processing.

dshanske:
Just going by the recommendation from the specification.

pfefferle:
Sure, but it makes the code way more complex and introduces problems we already fixed in the current version (with or without `www`, `http` vs `https`). The spec is also very vague on this topic:

> [...] For example, in an [ HTML5] document, the receiver should look for <a href="*">, , <video src="*"> and other similar links. [...]

dshanske:
Will issue a separate PR for the conditional issue then.

I can close this and move to the next thing on the list.

PR #230, opened 2019-10-04
Add support for outputting a list of homepage webmentions

tw2113:
Includes a template file for use wherever, as well as basic shortcode and widget support.


includes/class-webmention-homepage-widget.php
```diff
@@ -0,0 +1,74 @@
...
+	 */
+	public function __construct() {
+		parent::__construct(
+			'Webmention_Homepage_Webmentions_Widget',
+			esc_html__( 'Homepage Webmentions Widget', 'webmention' ),
+			[
+				'classname'   => 'webmention_homepage_webmentions_widget',
+				'description' => esc_html__( 'A widget that allows you to display your homepage webmentions.', 'webmention' ),
+			]
+		);
```
dshanske: WordPress naming standards suggest a correlation between filename and classname

includes/functions.php
```diff
@@ -357,3 +357,45 @@ function webmention_extract_urls( $content, $support_media_urls = false ) {
...
 	return array_filter( $urls );
 }
+
+/**
+ * Retrieve and display webmentions sent to the desingated homepage.
+ *
+ * @since NEXT
+ */
+function webmention_get_home_webmentions() {
+	$homepage = get_option( 'webmention_home_mentions', 0 );
```
dshanske: It should short circuit if 0 before the query

tw2113: Addressed in 1998135

includes/class-webmention-homepage-widget.php
tw2113: Addressed in 4fb9cee

tw2113:
Somehow, i broke things, so don't merge right now.

tw2113:
I probably did do some copy/paste and missed that, so good catch there.

PR #234, opened 2019-10-18
Notifications

dshanske:
This is the initial creation of support for improving comment email notifications for webmentions. As of now, it only displays a customized display for webmentions that is very similar to the default comment display.

As we add in more rendering over PHP-MF2, it will expand to show more. 

The version in Semantic Linkbacks uses the older filter, whereas this uses the newer hook to take control and either pass it to the core function, or a new local implementation based on comment type.


dshanske:
@pfefferle Any thoughts on this?

pfefferle:
Why do we have to copy nearly the whole functionality?

dshanske:
Because there is no easy way otherwise. I am thinking of submitting a core patch

pfefferle:
Why do the "normal" notification not work?

dshanske:
Because it doesn't display the relevant fields or handle custom comment types

includes/class-webmention-notifications.php
```diff
@@ -0,0 +1,426 @@
...
+			@wp_mail( $email, wp_specialchars_decode( $subject ), $notify_message, $message_headers );
+		}
+
+		if ( $switched_locale ) {
+			restore_previous_locale();
+		}
+		return true;
+	}
+
+
```
pfefferle: can you add the missing docblock?

includes/class-webmention-notifications.php
```diff
@@ -0,0 +1,426 @@
...
+				$message .= sprintf( __( 'URL: %s', 'webmention' ), $comment->comment_author_url ) . "\r\n";
+				/* translators: %s: comment text */
+				$message .= sprintf( __( 'Comment: %s', 'webmention' ), "\r\n" . get_comment_text( $comment ) ) . "\r\n\r\n";
+				break;
+		}
+		return apply_filters( 'webmention_comment_details', $message, $comment );
+
+	}
+
+
```
pfefferle: Docblock

includes/class-webmention-notifications.php
```diff
@@ -0,0 +1,426 @@
...
+					break;
+				case 'resend_moderation_email':
+					self::wp_notify_moderator( $comment_id );
+					break;
+			}
+		}
+		return $redirect_to;
+	}
+
+	/*
```
pfefferle: Docblocks start with /**

includes/class-webmention-notifications.php
```diff
@@ -0,0 +1,426 @@
...
+
+	}
+
+
+	public static function wp_new_comment_notify_postauthor( $comment ) {
+		$comment = get_comment( $comment );
+
+		$maybe_notify = get_option( 'comments_notify' );
+
+		/**
```
pfefferle: I am not sure, but isn't WordPress using `/*` for inline multiline docs?

PR #237, opened 2019-12-01
Add MF2 Parsing to Webmentions

dshanske:
For this first PR that includes php-mf2, the goal is to parse the MF2 in webmentions as opposed to Semantic Linkbacks. A separate PR(https://github.com/pfefferle/wordpress-semantic-linkbacks/pull/230) will pull that data in Semantic Linkbacks if it is available.

This makes several changes around loading files. It adds an autoloader that will load the webmention classes when they are first requested, which is, in the current classes, when their init functions are first added.

The MF2 parser is loaded as a dev dependency, and then a script inside composer that is run on update copies it to the lib directory along with the license and README files. This ensures the current version is bundled without the vendor directory, as we actually aren't using composer.

The MF2 Parser is only loaded inside the verify function. This is a change we made to the other plugins that call php-mf2 to minimize one version overriding another. 

At this time, it and DOMDocument are generated after text based verification. A future version may rearrange that to verify using the parsed version of the retrieved URL, but that was out of scope for this.

The final change involves, as part of the retrieval of a remote URL, adding several functions to do so, making a HEAD request, a GET request, verifying the status code and content type as separate functions. This allows for future expansion in these areas.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,8 @@
     "squizlabs/php_codesniffer": "^3.4",
     "phpcompatibility/php-compatibility": "^9.2",
     "wp-coding-standards/wpcs": "^2.0",
-    "phpcompatibility/phpcompatibility-wp": "^2.0.0"
+    "phpcompatibility/phpcompatibility-wp": "^2.0.0",
+    "mf2/mf2": "^0.4.6"
   },
   "scripts": {
     "test": [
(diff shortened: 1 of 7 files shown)


dshanske:
> This is a lot to review and a mix of different use cases, that makes it hard to give feedback. Why changing the license, btw?
> 
> Can we please start with only loading the libraries?

Where did I change the license? I included, in the directory where php-mf2 is stored, the php-mf2 license.

So, you want me to split the three things here?

1. Autoload

1. Refactoring the fetch and head functions

1. Loading php-mf2

pfefferle:
Sorry, it was the copied mf2 license. Shame on me... I shouldn‘t review mrs on the phone!

dshanske:
Made the mistake of doing it on phones as well

PR #239, opened 2019-12-02
Add mf2 library

pfefferle:
@dshanske what do you think?


pfefferle:
It prefixes the Mf2 lib, so we do not have to worry about naming collisions and/or different versions. We could also easily add the emoji lib by aaronpk.

pfefferle:
Oh god... this requires a whole lot of refactoring... will re-check later today...

dshanske:
A bit more complicated than what I did. Will wait for you to finish the refactoring

dshanske:
I'm not against going up to PHP5.6. I had just waited until we did something that needed it. But considering core won't allow below 5.6 anymore, isn't worth maintaining the compatibility issue in support of continuing to move forward.

pfefferle:
@dshanske should work now. The result is nearly the same as in your mr, but it adds `Webmention` as the main namespace, so that we do not run into naming and version issues.

dshanske:
@pfefferle But the mf2 isn't called yet. Do we want to merge it then add it actually being called?

pfefferle:
Yes, let’s work on the autoload next and after that on the parsing thingy

dshanske:
Then this PR is approved. I will set up a separate PR for adding the mf2 json to the comment data, then a separate one for the fetch refactoring I already wrote, then we'll start building actual usage of the mf2 data

dshanske:
For parsing, will start simple, by the way, with h-card parsing for author name and image, and move to comment types later

pfefferle:
Can we discuss the comment data thing? Is it a good idea to have it as part of the comment?

pfefferle:
I would also prefer to avoid the jf2 stuff, because it felt very hacky working on that in the mf2-feed plugin.

dshanske:
@pfefferle That is just passing it along to other functions, it never hits the database. As for jf2, only if that comment type is returned.

PR #245, opened 2019-12-25
Utilize Parsing

dshanske:
In the previous merge, we added PHP-MF2 into webmentions. In this PR, the system actually parses MF2 and passes it and the DOMDocument used to generate it into the $commendata array. The DOMDocument can be reused to parse title and other functions if there is no MF2 discovered, so there is no need to redo it.

In https://github.com/pfefferle/wordpress-semantic-linkbacks/pull/230 - Semantic Linkbacks looks for that same parsed MF2 and uses it if it is there. 

This also adds the filter discussed to pass in an alternative DOMDocument class.

### DIFF
--- a/includes/class-webmention-receiver.php
+++ b/includes/class-webmention-receiver.php
@@ -553,7 +553,7 @@ class Webmention_Receiver {
 
 		// If this is an mf2 object store
 		if ( 'application/mf2+json' === $content_type ) {
-			$commentdata['remote_source_mf2'] = json_decode( $remote_source_original );
+			$commentdata['remote_source_mf2'] = json_decode( $remote_source_original, true );
 		}
 		if ( 'text/html' === $content_type ) {
 			// Pass the DOMDocument as it can be used to add additional properties should MF2 parsing not yield them
--- a/includes/class-webmention-receiver.php
+++ b/includes/class-webmention-receiver.php
@@ -1,49 +1,49 @@
 <?php
 /**
- * Webmention Receiver Class
+ * Webmention Receiver Class.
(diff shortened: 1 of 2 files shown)


includes/class-webmention-receiver.php
```diff
@@ -554,6 +554,21 @@ public static function webmention_verify( $data ) {
 		$content_type  = wp_remote_retrieve_header( $response, 'Content-Type' );
 		$commentdata   = compact( 'remote_source', 'remote_source_original', 'content_type' );
 
+		// If this is an mf2 object store
+		if ( 'application/mf2+json' === $content_type ) {
+			$commentdata['remote_source_mf2'] = json_decode( $remote_source_original, true );
```
pfefferle, line 559: will this be saved to the db? should we use an underscore as prefix, to make clear that it is a temp "attribute"?

dshanske: We don't for remote_source or remote_source_original, which are derived from pingbacks doing the same thing.

pfefferle: It was not meant as a critic on this specific change. This was more a general question, but on a real world example. Why I asked: I think it is easier to understand/review the code when it is clear if the attribute is saved or not!?!

dshanske: I was just advising why I didn't. But at the moment, it isn't saved. I suppose we might if we ever tried for salmentions, but that would only need to save some basic comment data.

dshanske:
@pfefferle I think this we could possibly merge this.

PR #247, opened 2019-12-28
Initial Commit of MF2 Handling Code

dshanske:
This code adds the MF2 handling class, which includes helper functions adapted from https://github.com/barnabywalters/php-mf-cleaner to simplify the code. It also adds functions to get the representative author and item and return those.

For now, it includes some of the functionality from the MF2 Handling class in Semantic Linkbacks, but not all yet.

The goal of the code is to strip out everything but the single item that is the source of the mention, enhancing it by adding in the representative author if not present in that item. 

In order to do debugging on this, if someone advises that the system is not handling a site correctly, which we have numerous examples of in previously filed issues, I added in an endpoint and a debug form on the settings page that returns the output of this function.

It also splits off the fetch and head functions in the receiver so they can be called for the debugger.


dshanske:
This PR does not include the check to only return h-entry. It has been suggested that as there is no explicit reason to allow everything, so should include that until there is.

dshanske:
@pfefferle Do you think this is the right direction?

pfefferle:
The plugin-idea is very complex and it needs a lot of work to interoperate with the other plugins, so I am not sure if the plugin in general is the "right direction".

dshanske:
I know we want to improve webmention functionality as a goal. What is the solution then?

dshanske:
I know the lack of movement in Core is also an issue.

dshanske:
The alternative is to freeze both plugins and launch the replacement as a totally new project.

pfefferle:
sorry I was in the wrong context! I thought the question was for the mf2-feed plugin! I will have a look at the code!

includes/class-webmention-mf2-handler.php
```diff
@@ -0,0 +1,389 @@
+<?php
+/**
+ * takes MF2 JSON and processes it
+ *
+ */
+class Webmention_MF2_Handler {
```
pfefferle: Does it make sense to have an object here instead of a simple class? I only use classes for stateless libs.

includes/class-webmention-mf2-handler.php
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	/**
+	 * Verifies if property named $propname is in array $mf.
+	 *
+	 * @param array    $mf
+	 * @param $propname
+	 * @return bool
+	 */
+	public static function has_prop( array $mf, $propname ) {
```
pfefferle: Most of the functions have full/representative names, so I would prefer `has_property`.

includes/class-webmention-mf2-handler.php
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	/**
+	 *  Return an array of properties, and may contain plaintext content
+	 *
+	 * @param array       $mf
+	 * @param array       $properties
+	 * @return null|array
+	 */
+	public static function get_prop_array( array $mf, $properties, $args = null ) {
```
pfefferle: `get_property_array`

includes/class-webmention-receiver.php
```diff
@@ -465,49 +455,144 @@ public static function webmention_verify( $data ) {
...
+	public static function remote_arguments() {
+
+		$wp_version = get_bloginfo( 'version' );
+		$user_agent = apply_filters( 'http_headers_useragent', 'WordPress/' . $wp_version . '; ' . get_bloginfo( 'url' ) );
+		$args       = array(
+			'timeout'             => 100,
+			'limit_response_size' => 153600,
+			'redirection'         => 20,
+			'headers'             => array(
+				'Accept' => 'application/mf2+json, text/html, text/plain',
```
pfefferle: I would love to add `application/mf2+json` as a "beta" opt-in setting. I think, for now, the html representations are much more detailed than any `mf2` feeds.

includes/class-webmention-mf2-handler.php
dshanske: This is based a similar class in Semantic Linkbacks...but  so far it is just helping finding of representative author and item, stripping everything else. It could be an object...but this is more a stateless libs, as you described.

includes/class-webmention-mf2-handler.php
dshanske: That's fine. Some of these functions are based on the old mf-cleaner repo, so the names are inherited from that. Will adapt.

includes/class-webmention-receiver.php
dshanske: This was based on the fact Semantic Linkbacks used mf2+json.  But if you look at the priority, it will only offer it if the endpoint returns it. So far, only @aaronpk does as far as I know, so not an issue to drop it for now.

dshanske:
Agree on the unit tests. Tests need to see if for different types, if it returns what it should...a single item and author.

dshanske:
@pfefferle Do you want me to work on unit tests for this PR or work on that for a separate PR? As of now, this code is not automatically triggered by anything, as I planned that to be another PR as well.

dshanske:
@pfefferle Re the question above... should I work on adding unit tests here, or wait to a separate PR?

dshanske:
@pfefferle Asking again... merge then unit tests or other way round?

dshanske:
Didn't realize you had corrected some my docblocks....had a conflict. Also made some minor fixes.

@pfefferle Just to remind you, this code adds functions for handling mf2 json, it also helps identify the representative author and item.

dshanske:
It also adds a debugger to look at that data that appears when WP_DEBUG is enabled.

PR #249, opened 2020-01-01
Cache in cases where stored avatar is a gravatar

### LINKED ISSUE #248: New Webmention Avatar Handling Code Issue if MF2 Parser returns a gravatar
I commented out the Semantic Linkbacks class load on my website to test https://github.com/pfefferle/wordpress-semantic-linkbacks/issues/231 reported by @edent and started looking to see if I had a gravatar problem. I still do. 

So, here is the issue. The logic checks to see if the URL is a gravatar. But it also checks to see if there is an email set, on the assumption that all gravatars are email lookups. So, when people return a gravatar as part of parsed MF2, because there is no email, it falsely overrides it.

So, I am going to submit a PR for this scenario.

dshanske:
Fixes #248 

This moves the checking for whether an email exists from the anoynmous code into the check_gravatar function. It also passes the arguments array, optionally in and checks to see if the URL passed in it is a gravatar URL. If so, it extracts the hash and runs the same check.

This should fix the problem.

### DIFF
--- a/includes/class-webmention-avatar-handler.php
+++ b/includes/class-webmention-avatar-handler.php
@@ -83,8 +83,17 @@ class Webmention_Avatar_Handler {
 	 *
 	 * @return boolean
 	 */
-	public static function check_gravatar( $comment ) {
-		$hash  = md5( strtolower( trim( $comment->comment_author_email ) ) );
+	public static function check_gravatar( $comment, $args = null ) {
+		if ( ! empty( $comment->comment_author_email ) ) {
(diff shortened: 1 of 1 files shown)


dshanske:
@pfefferle Suggest we psuh this along with the related Semantic Linkbacks load issue to wordpress.org to improve the experience for those affected.

dshanske:
@pfefferle Would you have a chance to push this and the related Semantic Linkbacks fix?

PR #254, opened 2020-01-16
Re-set comment approval status for always allowed domains

jeremyfelt:
Even after the `comment_approved` key is set to `1` in the comment data, the comment goes through `wp_new_comment()` for storage. This in turn runs the comment through `wp_allow_comment()` which checks the comment with `check_comment()` and causes the `comment_approved` key to be set back to `0` if any kind of comment moderation is enabled on the site.

This change uses the available `pre_comment_approved` filter to check if the comment is a webmention previously approved based on its domain name and then returns `1` as expected.

### DIFF
--- a/includes/class-webmention-receiver.php
+++ b/includes/class-webmention-receiver.php
@@ -35,6 +35,8 @@ class Webmention_Receiver {
 
 		add_filter( 'webmention_comment_data', array( 'Webmention_Receiver', 'auto_approve' ), 13, 1 );
 
+		add_filter( 'pre_comment_approved', array( 'Webmention_Receiver', 'set_comment_approved' ), 10, 2 );
+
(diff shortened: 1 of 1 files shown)


dshanske:
@jeremyfelt Thanks for this. Want to look at it more closely, but that explains why it stopped working. They made changes to the allow comment a few versions ago.

dshanske:
@jeremyfelt Why not attach the auto approve function to the pre_comment_approved hook instead?

jeremyfelt:
> Why not attach the auto approve function to the pre_comment_approved hook instead?

Good question! I think I was just heads down on finding a way through the existing pattern and didn't think to go back and re-hook the whole thing. I'll take a look at #255 as I'm guessing it's the better option.

jeremyfelt:
Closing in favor of #255.

Issue #252, opened 2020-01-14
Change "approve & whitelist" to be inclusive and more reader friendly


jeremyfelt:
Howdy. Thanks for the excellent plugin! I've **really** been enjoying playing around with Webmentions finally.

I'd like to propose changing the "Approve & Whitelist" text below each pending comment to something [more inclusive](https://medium.com/@bronAtNuna/language-matters-7fced5d78f8d) and reader friendly.

"Approve and Allowlist" doesn't really read right, but "Approve and always allow" or "always allow" seem like decent alternatives. In addition to being inclusive, I think that an average blog owner would be more confident knowing what "always allow" means rather than "whitelist".

It appears that the language can change throughout as the option stored in the database is `webmention_approve_domains`, which does not cause any back-compat issues. It may be that the function, `transition_to_whitelist` needs to remain as is in case anyone had been removing that filter through another plugin.

I'm happy to submit a PR with changes if this fits. :)

dshanske:
No objection. I thought of doing this when the feature was working better, as it seems to be malfunctioning at the moment.

PR #255, opened 2020-01-18
Fix Comment Approval Alternate

dshanske:
This addresses #252 and is an alternative to #254. Props to @jeremyfelt who identified that the issue is the wp_allow_comment function interfering even though the comment_approved is set.

Now, the WEBMENTION_COMMENT_APPROVE flag has been there for years, even before the domain approval list. So, something else we changed created this situation. Either way, moving both checks to wp_allow_comment seems more consistent with the flow of adding a comment than setting it earlier.

At the same time, addressed the issue re the term whitelist. I did not change the function name, as it might be called elsewhere, but did change everything else.

Also bumped version to 4.0.3 for easy release.


jeremyfelt:
👍 I tested this out locally and it's working as expected.

dshanske:
@jeremyfelt Do you want to close your PR then and we'll wait for review?

dshanske:
@jeremyfelt FYI, @pfefferle and I are the only approved reviewers right now.

PR #257, opened 2020-02-11
Support deployment by action

dshanske:
I want to deploy 4.0.3 which is the current version in the github repo, but not on wordpress.org. This has the comment type avatar fix and the comment_approved fix.

I'm adding the Github deploy functionality added elsewhere. You already had the asset and readme update functionality set up. So if after you approve, we should just have to push the 4.0.3 tag and it will deploy.


dshanske:
@pfefferle If you have time for this today, I can push to WordPress.org I think this will resolve the avatar support complaint there.

dshanske:
https://wordpress.org/support/topic/avatars-broken-in-dashboard-2/

PR #262, opened 2020-04-28
Local avatar cache

pfefferle:


pfefferle:
@snarfed @dshanske @tw2113 @chrisaldrich any feedback?

Currently I save the local path, the same way we save the remote image path... Is this a good idea? Would it be better to only have a flag `avatar_cached` or something like that?

dshanske:
> @snarfed @dshanske @tw2113 @chrisaldrich any feedback?
> 
> Currently I save the local path, the same way we save the remote image path... Is this a good idea? Would it be better to only have a flag `avatar_cached` or something like that?

The only issue would be site migration, where the URL would change, but that's a problem search and replace plugins could solve.

dshanske:
Do we ever want to expire these? Looking at the code. Once an avatar is set, it seems set for life.

dshanske:
Thinking about that...it would just get overwritten during the next webmention from the same user/domain.

pfefferle:
> Do we ever want to expire these? Looking at the code. Once an avatar is set, it seems set for life.

A newer version will replace the old version and that will update all old comments/reactions with the new image.

pfefferle:
> The only issue would be site migration, where the URL would change, but that's a problem search and replace plugins could solve.

But why? If the URL changes, it is like a new user. The old comments keep the old avatar, the new comments will have the new. I think this is an edge case we can think about later...

pfefferle:
@dshanske do you prefer the URL thing or the flag, that there is a cache item?

dshanske:
> > The only issue would be site migration, where the URL would change, but that's a problem search and replace plugins could solve.
> 
> But why? If the URL changes, it is like a new user. The old comments keep the old avatar, the new comments will have the new. I think this is an edge case we can think about later...

I meant if the host site changes. So, if my site changes and suddenly, the URLs don't resolve.

pfefferle:
Ah, I see! Ok, then I am with you... Should be handled by search/replace.

dshanske:
> @dshanske do you prefer the URL thing or the flag, that there is a cache item?

Well, the way I set up the avatar property is that it can be a numeric, referencing a media attachment ID, or a URL. How about a value of 0 ? Which would trigger a lookup in the cache directory?

pfefferle:
> Well, the way I set up the avatar property is that it can be a numeric, referencing a media attachment ID, or a URL. How about a value of 0 ? Which would trigger a lookup in the cache directory?

I would not prefer `0` because it tastes like `false` or `null` but we can use an enum, like `cache` or something similar.

dshanske:
> > Well, the way I set up the avatar property is that it can be a numeric, referencing a media attachment ID, or a URL. How about a value of 0 ? Which would trigger a lookup in the cache directory?
> 
> I would not prefer `0` because it tastes like `false` or `null` but we can use an enum, like `cache` or something similar.

Works for me.

mundschenk-at:
From an interoperatibility point of view, you should add filter hooks to the constants so that plugins that do their own caching/resizing of avatars can request a higher resolution base image. Also, if you never delete/recreate your images, I'm not sure it counts as "caching" from a legal perspective.

pfefferle:
@mundschenk-at this is an issue of twitter & co. because their image URLs are not consitent. That means if a user changes his avatar the URL is broken.

florianbrinkmann:
👋 what is missing to get this PR merged? I just installed the webmention plugin yesterday and am happy to see a PR for local avatars, because I would prefer them over the Twitter image links (because of privacy, and because they are blocked by default in Firefox, at least the dev version) :)

pfefferle:
It needs some code to migrate old avatars and a setting to enable disable it (or at least a feature switch using a const). And I am not completely happy with the way I store the images. Do you have some feedback?

dshanske:
I'll look it over and see if I have any suggestions, but I know this is a tough one

florianbrinkmann:
> It needs some code to migrate old avatars and a setting to enable disable it (or at least a feature switch using a const). And I am not completely happy with the way I store the images. Do you have some feedback?

I will take a look at it next week and see if I can help somehow.

dshanske:
I rebased the branch. I think I have a few things I might try here.

dshanske:
I think I want to remove the filehandle from the stored meta and have the URL built in get_avatar.

mundschenk-at:
> I think I want to remove the filehandle from the stored meta and have the URL built in get_avatar.

As mentioned in https://github.com/pfefferle/wordpress-webmention/pull/262#issuecomment-644645486, it would be nice if you could add filter hooks to to allow the image to be modified (e.g. scaled) by other plugins).

dshanske:
@pfefferle Made some modifications here. Switched to using the wp_upload_dir function over the constant. Added a function to convert back to the filepath from the URL and to delete the avatars.  I also tested it by sending a local webmention. It did store the gravatar photo, which is what I used for the test.

Now, the storage is currently at hostname/username which isn't a bad methodology. We could also go for a single directory and a hashed filename, I suppose. What makes sense?

To @mundschenk-at, it is more of a 'store' than a cache. The images only refresh when the webmention is refreshed.

dshanske:
I was thinking about the turning off functionality. I think the turning off functionality may be not storing any image data. There's an existing setting to show avatars for webmentions. We might as well change it to store and show.

mundschenk-at:
@dshanske Yes, but making things filterable is good practice (even if in this case, my "Legacy Icon" handler could probably deal with the current implementation).

dshanske:
@mundschenk-at Agree about the filters. Figuring out how it should work first. We still haven't addressed purging and a few other items..

pfefferle:
@mundschenk-at do you have suggestions? Perhaps we can directly add filters to support your avatar cache plugin as extension.

mundschenk-at:
It's probably easier the other way around (i.e. Avatar Privacy add's support for integrating Webmentions). That said, adding filters for the URLs returned to (`pre_`)`get_avatar_data` should probably do the trick. (Though on reflection, I think that those will probably not be strictly necessary, unhooking the avatar integration and calling the functions directly would probably also work.)

florianbrinkmann:
Do I understand that correctly: Currently, an avatar of user X would be stored without checking for new versions until there is a new webmention from that user X, then the plugin would fetch the avatar again and update it? That means, when I have only a single webmention from a user, and that user changes the Twitter image, I will never get the updated image on my site?

> There's an existing setting to show avatars for webmentions. We might as well change it to store and show.

I think that would be a good option, always store them locally when the avatar feature is enabled, enhanced privacy by default :) Maybe a filter could be added to disable that when a user really needs/wants to load them from the social network.

@mundschenk-at if you are thinking about adding webmention avatar support to your Avatar Privacy, a filter would be needed to disable local storage in the webmention plugin, right? To not store two versions of the avatars.

pfefferle:
> Do I understand that correctly: Currently, an avatar of user X would be stored without checking for new versions until there is a new webmention from that user X, then the plugin would fetch the avatar again and update it? That means, when I have only a single webmention from a user, and that user changes the Twitter image, I will never get the updated image on my site?

Yes, but this is an issue of Twitter... Twitter has no persistant Avatar URLs, so if you change the Avatar, you have a new Avatar URL. So we only know about the new URL when there is a new Webmention.

florianbrinkmann:
But couldn’t we check the user information via the Twitter API or something like that and get the current avatar URL from there?

pfefferle:
Sure, but this is very Twitter specific and not a generic feature of Webmentions... In the end, it only works because of Brid.gy! So yes, we can add some proprietary optimisations, but I would like to have a generic best practice first ;)

@florianbrinkmann what do you think about the code (except the twitter thingy? ;) )

pfefferle:
@florianbrinkmann it is also not that easy to handle such an "upate" process... perhaps as a cron, but we have to save the time, we tried to update the icon last...

mundschenk-at:
@florianbrinkmann I have added a generic "remote image" support. I have not tested it with the implementation in this PR, but but I'd probably treat it as a "remote storage" for purposes of my plugin, not replace it with my own mechanism, but cache resized variations.

pfefferle:
@florianbrinkmann but twitter is also a reason, that it is not that easy to use @mundschenk-at plugin because, as I said, the twitter URL changes and it we have to at least invest some time to work around that issue.

florianbrinkmann:
> Sure, but this is very Twitter specific and not a generic feature of Webmentions...

Yes, I didn’t have that in my mind, you are right.

> So yes, we can add some proprietary optimisations, but I would like to have a generic best practice first ;)

Yes, you are right, that makes sense :)

> @florianbrinkmann what do you think about the code (except the twitter thingy? ;) )

At the first glance, it looks good for me, I will try to test the PR at the weekend.

> @florianbrinkmann I have added a generic "remote image" support. I have not tested it with the implementation in this PR, but but I'd probably treat it as a "remote storage" for purposes of my plugin, not replace it with my own mechanism, but cache resized variations.

Ah okay.

florianbrinkmann:
I just uploaded the code to my site and tried to resend a Twitter webmention via Bridgy, but I get the following error:

```
PHP Fatal error:  Uncaught Error: __clone method called on non-object in /…/wp-content/plugins/webmention/includes/entities/class-webmention-item.php:271
Stack trace:
#0 /…/wp-content/plugins/webmention/includes/class-webmention-receiver.php(626): Webmention_Item->to_commentdata_array()
#1 /…/wp-includes/class-wp-hook.php(287): Webmention_Receiver::default_commentdata(Array)
#2 /…/wp-includes/plugin.php(206): WP_Hook->apply_filters(Array, Array)
#3 /…/wp-content/plugins/webmention/includes/class-webmention-receiver.php(339): apply_filters('webmention_comm...', Array)
#4 /…/wp-includes/rest-api/class-wp-rest-server.php(1050): Webmention_Receiver::post(Object(WP_REST_Request))
#5 /…/wp-includes/rest-api/class-wp-rest-server.php(376): WP_REST_Server->dispatch(Object(WP_REST_Request))
#6 /…/wp-includes/rest-api.php(339): WP_REST_Serve in /…/wp-content/plugins/webmention/includes/entities/class-webmention-item.php on line 271```

pfefferle:
Have you used the latest master/trunk version? I would not recommend that yet, because we are working on merging the semantic linkbacks plugin into the webmentions plugins and there might be still some glitches.

florianbrinkmann:
I used the `local-avatar-cache` branch.

pfefferle:
Oh shit, I might merged some parts of the master into the branch. I will try to fix that.

dshanske:
@pfefferle Try merging the fixes I made in the other branch?

dshanske:
@pfefferle It might have been me when I tried working on the code

pfefferle:
hmmm... not sure... would be nice, to have this also in 4.x...

PR #263, opened 2020-05-05
Make admin stuff start from a hook

szepeviktor:
pluggable :)

### DIFF
--- a/webmention.php
+++ b/webmention.php
@@ -25,13 +25,18 @@ defined( 'WEBMENTION_PROCESS_TYPE' ) || define( 'WEBMENTION_PROCESS_TYPE', WEBME
 
 defined( 'WEBMENTION_VOUCH' ) || define( 'WEBMENTION_VOUCH', false );
 
-// initialize admin settings.
-require_once dirname( __FILE__ ) . '/includes/class-webmention-admin.php';
-add_action( 'admin_init', array( 'Webmention_Admin', 'init' ) );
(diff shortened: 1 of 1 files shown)


pfefferle:
I see! 

I like the example you referenced, can you help restructuring the plugin a bit?

szepeviktor:
Setting go from constants to Hooks

```
// defined( 'WEBMENTION_ALWAYS_SHOW_HEADERS' ) || define( 'WEBMENTION_ALWAYS_SHOW_HEADERS', 0 );

// In should_show_headers()
if ( apply_filters( 'webmention_always_how_headers', false ) )
```

Class loading will be done by Composer autoload

```
// require_once dirname( __FILE__ ) . '/includes/class-webmention-admin.php';

// NOTHING :)
```

Hooked functions go into a class

```
// function webmention_comment_form() {
```

Do you agree?

dshanske:
I personally am reluctant to use composer autoload because it creates a large plugin dependency. I usually limit composer to a development dependency myself.

szepeviktor:
> it creates a large plugin dependency.

Yes, currently with `composer/installers` it is +500 KB
But distributing only the autoload:

```
$ du -sh vendor/
56K     vendor/
```

:)

szepeviktor:
@pfefferle Sad End?

pfefferle:
I messed up with the branching... and github seems to close some merge requests because of this :(

pfefferle:
I will work throught the MRs to try to fix them... I even can't reopen them! argh!

szepeviktor:
Thank you.

PR #264, opened 2020-06-14
Add functionality to register a webmention comment type

dshanske:
The idea here is a simple registration class for the webmention comment types based on the register_post_type function. 

So far, it incorporates the strings used for the various internal labels and menus, but eventually would add others as needed.

Essentially, it is an alternative to adding strings in five different places when adding a new type. And it will make it easier to add additional configuration options as needed.

If WordPress ever adds its own comment type registration, this would replace that.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1,4 +1,40 @@
 <?php
+
+/**
+ * Registers a webmention comment type.
+ *
+ *
+ * @param string $comment_type Key for comment type.
+ * @param array $args Arguments.
+ *
+ * @return Webmention_Comment_Type The registered webmention comment type.
+ */
+function register_webmention_comment_type( $comment_type, $args = array() ) {
+	global $webmention_comment_types;
+
+	if ( ! is_array( $webmention_comment_types ) ) {
+		$webmention_comment_types = array();
+	}
+
+	// Sanitize comment type name.
+	$comment_type = sanitize_key( $comment_type );
+
+	$comment_type_object = new Webmention_Comment_Type( $comment_type, $args );
+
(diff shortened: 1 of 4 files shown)


dshanske:
The other thing I want to do with this in future is allow for custom comment handlers to be registered for each type. We had a template request for that in Semantic Linkbacks. It would be nice to accommodate it.

pfefferle:
I thought we wanted to make the current subtypes to be the main comment types? So reply, repost, ... instead of Webmention?

dshanske:
Yes, this is the start of the code for that

dshanske:
So, this code would register reply, like, etc.

includes/functions.php
```diff
@@ -401,27 +437,37 @@ function is_webmention_comment_type( $comment ) {
...
-			break;
-		default:
-			$name = __( 'Response', 'webmention' );
+
+	$name = null;
+
+	if ( is_array( $webmention_comment_types ) && array_key_exists( $type, $webmention_comment_types ) ) {
+		$name = $webmention_comment_types[ $type ]->singular;
 	}
+	if ( ! $name ) {
```
pfefferle, line 452: shouldn't this be part of the new class?

pfefferle:
where should the default types be registered? maybe we should also add a class & rel list, to also include https://github.com/pfefferle/wordpress-semantic-linkbacks/blob/master/includes/class-linkbacks-mf2-handler.php#L23

dshanske:
@pfefferle The class and rel list was something I planned for additions. This PR is just the basic design.

Issue #195, opened 2019-05-06
New Fetching Functionality


dshanske:
Working on breaking out the functionality that retrieves webmentions, verifies them, and passes the source to be used to generate stored properties as well as presentation.

Using this issue to take notes and get feedback on how that might work.

1. Do a head request on the url. Check that this is an approved mime type... exclude media content as media cannot be the source of a webmention.

1. Check response codes...if not 200 series, accounting for redirects, then fail. Check to see if existing webmention matches source/target to account for deletion/etc.

1. Normalize URLs to assist in text matching.

1. Retrieve source. Address different media types differently. Specifically, for html verify that the url is part of a link. For json, a property, etc. For plain text and unrecognized types, existing text search.

1. Once verification is successful, pass an array containing the original source in key remote_source_original, matching core.

dshanske:
1. If HTML, then a Domdocument object should exist already, having been used to handle looking for URLs on href and src properties only. PHP-MF2 can accept that... saving the system the need to generate a new one 

1. For all other media types, convert output to a mf2 representation if possible.

1. Generate an mf2 json representation of the document by using the existing Domdocument object. Pass this using the remote_source key.

1. Check to see if the mf2 object is empty or is missing key properties...title, summary/content. If missing, try to backfill these using page title and meta tags.

1. Add a filter at this phase that allows the parsed remote source to be filtered. A plugin could be added to convert other markup to mf2 at this point if someone wanted to write in future. 

1. Return the content type, response code, as well as the original source, parsed source, and other relevant properties. The function that called this functionality can then call presentation and storage functions.

dshanske:
To support private posts in future, the failure on retrieval needs to support 401 with a WWW-Authenticate header with a hook to allow for auth support to be added later.

PR #265, opened 2020-06-15
Create class-webmention-request.php

pfefferle:
I would love to simplify things, because the code is getting larger and larger.

I would love to have some very simple methods, to return the HTML and the correct parsed MF2 or directly the prefilled comment.

Something like:

```PHP
$request = new Webmention_Request();
$response = $request-fetch( $url );

if ( is_wp_error( $response ) ) {
	return $response;
}
$mf2_handler = new Webmention_Mf2_Handler();
$mf2_handler->parse( $response->get_body(), $url );
$mf2_handler->prefill_comment();
```

In this case it is very easy to use the code for the admin debug endpoint, for the direct parsing and the scheduled parsing.

I would love to work on this one first, so that we can use the request object and adapt the changes for the #247 change.

Perhaps we could also add the send webmention code to this class.

@dshanske what do you think?

@dshanske please commit directly to this branch if you like the idea.

### DIFF
--- a/includes/class-webmention-request.php
+++ b/includes/class-webmention-request.php
@@ -37,7 +37,7 @@ class Webmention_Request_Handler {
 			$response = wp_remote_get( $url, $args );
 		}
 		if ( is_wp_error( $response ) ) {
-				return $response;
+			return $response;
 		}
 		$check = $this->check_response_code( wp_remote_retrieve_response_code( $response ) );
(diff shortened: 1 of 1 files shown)


dshanske:
I brainstormed this in https://github.com/pfefferle/wordpress-webmention/issues/195

PR #269, opened 2020-06-17
Requests

dshanske:
This integrates the requests object into the webmention verification project. To execute the idea of moving the verification functionality in the future, need to wait for the parsing handlers...as it makes sense to rewrite that then.


includes/class-webmention-receiver.php
```diff
@@ -557,9 +484,11 @@ public static function webmention_verify( $data ) {
 			include_once ABSPATH . 'wp-includes/kses.php';
 		}
 
-		$remote_source = wp_kses_post( $remote_source_original );
-		$content_type  = wp_remote_retrieve_header( $response, 'Content-Type' );
-		$commentdata   = compact( 'remote_source', 'remote_source_original', 'content_type' );
+		$commentdata = array(
+			'content_type'           => $request->get_content_type(),
+			'remote_source_original' => $request->get_body(),
+			'remote_source'          => $request->sanitize_html( $request->get_body() ),
```
pfefferle, line 490: `sanitize_html` is a static function. What about moving it to the `functions.php' as a global function?

pfefferle:
This looks so much cleaner! very nice!

includes/class-webmention-receiver.php
dshanske: Would need a prefix, but no objection.

PR #271, opened 2020-06-18
Webmention Handler Class

dshanske:
Still brainstorming, but this abstract class is the start of the mf2, ogp, etc handler.


pfefferle:
I think we should define the data format, so that we will be able to check if/how much of the attributes are set by the format parser.

Something like `Webmention_Entity` (could be an other name), that represents the properties we need to build the comment. The Entity should have at least two functions, one to check if it is a valid comment (has all needed properties) and one to export to commentdata.

@dshanske what do you think?

pfefferle:
Something like (pseudocode):

```PHP
$data = new Webmention_Entity();
$mf_parser = new Webmention_Mf2_Handler();

$handlers = apply_filters( 'webmention_entities', array( $mf_parser ) ); // we could also add a register style like for comment-types

foreach ( $handlers as $handler ) {
    $data = $handler->parse( $request, $data );

    if ( $data->is_complete() ) {
          exit;
    } else {
          continue;
    }
}

$commentdata = $data->to_commentdata();
```

dshanske:
How would you feel if the data format was jf2? It's probably the simplest. We can use the principles of jf2 common for Microsub https://indieweb.org/Microsub-spec#Posts  as it varies slightly from https://jf2.spec.indieweb.org/ So no matter what the input format is, we output that way.

We can add in the discovered type of comment, reply/like/etc as a property.

pfefferle:
How does JF2 display the content-type?

pfefferle:
It is not very handy, if it is:

```JSON
{
    "in-reply-to": "https://..."
}
```

I would prefer something like:

```JSON
{
    "_content_type": "reply"
}
```

pfefferle:
OK, so we should generate a `Webmention_Entity` (and optional a `Webmention_Author_Entity` ?) with the properties of the JF2-entry. All data, we do not know could be added as a `_raw` property and be saved to a comment-meta.

pfefferle:
To discuss: https://github.com/pfefferle/wordpress-webmention/pull/271/files#diff-8ad6cf649eb4d0935c31b70b4ad28ca3

dshanske:
Microsub does as you suggest, a _ prefixed property for detected post type

dshanske:
I like the idea of the item and author classes, prior art might be SimplePie

pfefferle:
Feel free to rename the files!

pfefferle:
Perhaps we should also create a handler folder and a (rest-)api folder?

dshanske:
More to do, but I started by adding some code for a handler for meta tags...haven't added the code to convert that to the entity properties yet. Had to add a DOMDocument property to the request, partly because all handlers are likely going to use it, and why reparse it for each handler? The code to generate the OGP array is based on the version I use in Parse This.

pfefferle:
I think we should generate the entity objects outside of the handler, so we can cascade it. Currently every handler builds his own entity. Should we add it to the constructor or also to the parser method?

pfefferle:
what about `wp_get_meta_tags` can we remove the function in favor of the handler class?

dshanske:
> what about `wp_get_meta_tags` can we remove the function in favor of the handler class?

Yes, I will once the meta handler is fully built.

dshanske:
The reason why I picked meta to do first is it is much simpler than the mf2 one, which means it is easier to develop the class structure so if it reveals any ideas about the base class or structure, it will be easieer.

pfefferle:
> The reason why I picked meta to do first is it is much simpler than the mf2 one, which means it is easier to develop the class structure so if it reveals any ideas about the base class or structure, it will be easieer.

I totally agree!

pfefferle:
@dshanske I think we should generate the entity objects outside of the handler, so we can cascade it. Currently every handler builds its own entity. Should we add it to the constructor or also to the parser method?

dshanske:
The Parser method, making it optional so it can be generated if null

pfefferle:
@dshanske should we generate another class to iterate through the handlers or should this be a global function?

pfefferle:
The `Webmention_Handler_Handler` 😄

dshanske:
I think a class makes more sense as it can hold the individual handler instances.

includes/handlers/class-webmention-handler-meta.php
```diff
@@ -0,0 +1,199 @@
...
+			$meta = self::add_property( $meta, $meta_name, $meta_value );
+		}
+
+		$meta['title'] = trim( $xpath->query( '//title' )->item( 0 )->textContent );
+		$meta          = self::parse_ogp( $meta );
+		if ( isset( $meta['og'] ) ) {
+			$meta['og'] = self::parse_ogp( $meta['og'] );
+		}
+
+		$this->webmention_item = new Webmention_Item();
```
pfefferle, line 45: this replaces the item if there is one or not. I would like to require the item as param.

includes/handlers/class-webmention-handler-meta.php
```diff
@@ -0,0 +1,199 @@
...
+
+		// Set raw data.
+		$this->webmention_item->set__raw( $meta );
+		$this->webmention_item->set__response_type = 'mention';
+		$item                                      = array();
+		// Start by looking at OGP.
+		if ( isset( $meta['og'] ) ) {
```
pfefferle, line 52: Can we split this huge function into smaller parts? For example `parse_meta()`, `parse_dc()` and `parse_ogp()` ?

includes/entities/class-webmention-item.php
```diff
@@ -0,0 +1,145 @@
...
+	 * @return void
+	 */
+	public function __call( $method, $params ) {
+		$var = strtolower( substr( $method, 4 ) );
+
+		if ( strncasecmp( $method, 'get', 3 ) === 0 ) {
+			return $this->$var;
+		}
+
+		if ( strncasecmp( $method, 'set', 3 ) === 0 ) {
```
pfefferle, line 117: should we add a check here, so that the property will not be overwritten, once set? first come first serve?

includes/handlers/class-webmention-handler-meta.php
dshanske: Will do

includes/entities/class-webmention-item.php
dshanske: How about the magic setter allows for a second parameter that would overwrite if true, but otherwise doesn't?

dshanske:
Added the basic classes for Meta/OGP, MF2, and JSON-LD. These are the three simplest ones to work with. I recently adopted JSON-LD as an option for Parse This because it is everywhere. So, now having the basis for three handlers, I can start testing the handoff between them. 

Also added in the option to send a true as a second parameter into the set function to allow it to override what is set.

dshanske:
So, far, the Meta Class is fairly complete. The JSON-LD and MF2 classes so far just handle the actual parsing, not the post-processing. I've found that the best order is MF2, which will give the richest return, followed by JSON-LD, then Meta.

dshanske:
The next class is the Handler itself, that iterates through the handlers and sets the final item.

includes/handlers/class-webmention-handler-meta.php
```diff
@@ -0,0 +1,208 @@
...
+		}
+
+		$item = array_filter( $item );
+		foreach ( $item as $key => $value ) {
+			$key = 'set_' . $key;
+			$this->webmention_item->$key( $value );
+		}
+	}
+
+	protected function ogp( $meta ) {
```
pfefferle, line 77: can we add a verb here? `check_ogp` or `parse_ogp` or `interpret_ogp` or ... to be more descriptive?

includes/handlers/class-webmention-handler-meta.php
```diff
@@ -0,0 +1,208 @@
...
+					} elseif ( isset( $meta['article']['modified'] ) ) {
+						$item['updated'] = new DateTimeImmutable( $meta['article']['modified'] );
+					}
+				}
+			}
+		}
+		return $item;
+	}
+
+	protected function dublin_core( $meta ) {
```
pfefferle, line 136: same here... can we add a verb?

includes/handlers/class-webmention-handler-meta.php
```diff
@@ -0,0 +1,208 @@
...
+		} elseif ( isset( $meta['datePublished'] ) ) {
+			$item['published'] = new DateTimeImmutable( $meta['datePublished'] );
+		}
+
+		if ( ! isset( $item['author'] ) ) {
+			$item['author'] = array( 'name' => $meta['author'] );
+		}
+
+		$item = array_filter( $item );
+		foreach ( $item as $key => $value ) {
```
pfefferle, line 71: why such a "complicated" usage of different array formats instead of directly add them to the item-object? It is very hard to verify it like this. It feels like it does things twice.

dshanske: I was trying to make it simpler to read, but didn't quite work out. I can redo it to touch the object...

includes/handlers/class-webmention-handler.php
```diff
@@ -0,0 +1,63 @@
...
+
+
+	/**
+	 * Iterate through a list of handlers and return an item.
+	 *
+	 * @return Webmention_Handler_Item
+	 *
+	 */
+	public function parse( $request ) {
+		$item = null;
```
pfefferle, line 50: If we want to cascade the data, we should add an initiated item... If every parser should only use his data, we can keep it like that.

dshanske: > If we want to cascade the data, we should add an initiated item... If every parser should only use his data, we can keep it like that.

Since we're passing the raw data, that may be the only property I'd want to pass all of back, for all of the triggered handlers.

dshanske:
@pfefferle My question is what makes a Webmention_Item complete? We want to try to get an author name at the least...but what else?

includes/entities/class-webmention-item.php
```diff
@@ -0,0 +1,155 @@
...
+
+	/**
+	 * The source URL.
+	 *
+	 * @var string
+	 */
+	protected $url;
+
+	// mabe also an entity
+	protected $author = array( 'type', 'name', 'url', 'photo' );
```
pfefferle, line 36: should we add an object here?

includes/handlers/class-webmention-handler-jsonld.php
```diff
@@ -0,0 +1,111 @@
...
+
+	/**
+	 * Takes a request object and parses it.
+	 *
+	 * @param Webmention_Request $request Request Object.
+	 * @param Webmention_Item $item A Parsed Item. If null, a new one will be created.
+	 * @return WP_Error|true Return error or true if successful.
+	 */
+	public function parse( $request, $item = null ) {
+		if ( $item instanceof Webmention_Item ) {
```
pfefferle, line 22: this is used by every parse function, I would like to refactor this. Should we really keep this as param?

includes/handlers/class-webmention-handler.php
```diff
@@ -0,0 +1,62 @@
...
+
+	protected $handlers = array();
+
+	/**
+	 * Must be instantiated with at least one handler.
+	 *
+	 * @param array|Webmention_Handler_Base $handler
+	 *
+	 */
+	public function __construct( $handler ) {
```
pfefferle, line 15: this is a bit confusing! why adding a handler with the constructor? I would prefer to add all bundled handlers by default.

includes/handlers/class-webmention-handler.php
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	/**
+	 * Insert a Handler at the front of the list
+	 *
+	 * @param Webmention_Handler_Base $handler
+	 *
+	 */
+	public function unshift( $handler ) {
+		array_unshift( $this->handlers[], $handler );
```
pfefferle, line 39: is this correct? not `array_unshift( $this->handlers, $handler );` instead?

includes/handlers/class-webmention-handler-meta.php
```diff
@@ -0,0 +1,224 @@
...
+		$dom  = clone $request->get_domdocument();
+		$meta = $this->parse_meta( $dom );
+
+		// Set raw data.
+		$this->webmention_item->set__raw( $meta );
+
+		// OGP has no concept of anything but mention so it is always a mention.
+		$this->webmention_item->set__response_type = 'mention';
+
+		$this->ogp_to_jf2( $meta );
```
pfefferle, line 36: this is a confusing naming, because it does not really transfer to jf2, but it fills the object attributes instead.

includes/handlers/class-webmention-handler-meta.php
```diff
@@ -0,0 +1,224 @@
...
+		$this->dublin_core_to_jf2( $meta );
+
+		// If Site Name is not set use domain name less www
+		if ( ! $this->webmention_item->has_site_name() && $this->webmention_item->has_url() ) {
+			$this->webmention_item->set__site_name( preg_replace( '/^www\./', '', wp_parse_url( $this->webmention_item->get_url(), PHP_URL_HOST ) ) );
+		}
+
+		$this->webmention_item->set_name( $meta['title'] );
+
+		if ( ! $this->webmention_item->has_published() ) {
```
pfefferle, line 46: we used an object, to not add these checks to the specific handler... it should be part of the setter.

includes/handlers/class-webmention-handler-meta.php
```diff
@@ -0,0 +1,224 @@
...
+			if ( isset( $meta['citation_date'] ) ) {
```
pfefferle, line 47: this could be also part of the setter, so we do not have to check this every time.

includes/handlers/class-webmention-handler-jsonld.php
```diff
@@ -0,0 +1,109 @@
...
+		$jsonld = array_filter( $jsonld );
+		if ( 1 === count( $jsonld ) && wp_is_numeric_array( $jsonld[0] ) ) {
+			$jsonld = $jsonld[0];
+		}
+
+		// Set raw data.
+		$this->webmention_item->set__raw( $jsonld );
+		$this->webmention_item->set__response_type = 'mention';
+
+		foreach ( $jsonld as $json ) {
```
pfefferle, line 44: there is no break, so it always checks all json-ld objects and uses the last one.

includes/handlers/class-webmention-handler-meta.php
pfefferle: needs to have a check. if OGP has all properties, there is no need to also check DC.

pfefferle:
@dshanske before having too many handlers... can we finish one handler and then work on the others, to have a best practice? what do you prefer? json-ld or meta?

includes/handlers/class-webmention-handler-meta.php
```diff
@@ -0,0 +1,224 @@
...
+		} elseif ( is_array( $array[ $key ] ) ) {
+			$array[ $key ][] = $value;
+		}
+		return $array;
+	}
+
+	/**
+	 * Tries to parse OGP Meta tags by hierarchy.
+	 */
+	protected function parse_ogp( $meta ) {
```
pfefferle, line 189: I think the parse prefix is only confusing, because it does not "parse" the dom document, but filters the meta array. It still feels like a lot of re-arranging of arrays.

includes/handlers/class-webmention-handler-meta.php
```diff
@@ -0,0 +1,224 @@
...
+			$meta_value = $tag->getAttribute( 'content' );
+			// Sanity check. $key is usually things like 'title', 'description', 'keywords', etc.
+			if ( strlen( $meta_name ) > 200 ) {
+				continue;
+			}
+			$meta = $this->add_property( $meta, $meta_name, $meta_value );
+		}
+
+		$meta['title'] = trim( $xpath->query( '//title' )->item( 0 )->textContent );
+		$meta          = $this->parse_ogp( $meta );
```
pfefferle, line 86: why parsed twice? first into `$meta` then into `$meta['og]`?

includes/handlers/class-webmention-handler-meta.php
```diff
@@ -0,0 +1,224 @@
...
+					}
+				}
+			}
+		}
+	}
+
+	protected function dublin_core_to_jf2( $meta ) {
+		$item = array();
+		// Then look at Dublin Core Properties
+		if ( isset( $meta['dc'] ) ) {
```
pfefferle, line 156: where is `$meta['dc']` set?

includes/handlers/class-webmention-handler-meta.php
pfefferle: the function seems to also "parse" DC and any other "namespaced" vocabulary, so I would use a more generic name.

includes/handlers/class-webmention-handler-meta.php
```diff
@@ -0,0 +1,224 @@
...
+			if ( strlen( $meta_name ) > 200 ) {
+				continue;
+			}
+			$meta = $this->add_property( $meta, $meta_name, $meta_value );
+		}
+
+		$meta['title'] = trim( $xpath->query( '//title' )->item( 0 )->textContent );
+		$meta          = $this->parse_ogp( $meta );
+		if ( isset( $meta['og'] ) ) {
+			$meta['og'] = $this->parse_ogp( $meta['og'] );
```
pfefferle, line 88: Does this really work? it seems to result in `$meta['og']['og']` or am I wrong? If `parse_ogp` also interprets `DC` it will also result in `$meta['og']['dc']`.

includes/handlers/class-webmention-handler-jsonld.php
```diff
@@ -0,0 +1,109 @@
...
+				}
+
+				$this->webmention_item->set_name( $json['headline'] );
+				$this->webmention_item->set_name( $json['name'] );
+				$this->webmention_item->set_summary( $json['description'] );
+				$this->webmention_item->set_category( $json['keywords'] );
+
+				if ( isset( $json['articleBody'] ) ) {
+					$html            = webmention_sanitize_html( $json['articleBody'] );
+					$json['content'] = array(
```
pfefferle, line 63: we defined `$content` as string.

includes/handlers/class-webmention-handler-meta.php
```diff
@@ -0,0 +1,224 @@
...
+			}
+			if ( ! $meta_name ) {
+				$meta_name = $tag->getAttribute( 'itemprop' );
+			}
+			$meta_value = $tag->getAttribute( 'content' );
+			// Sanity check. $key is usually things like 'title', 'description', 'keywords', etc.
+			if ( strlen( $meta_name ) > 200 ) {
+				continue;
+			}
+			$meta = $this->add_property( $meta, $meta_name, $meta_value );
```
pfefferle, line 82: why not try to directly add to the item? we know the meta names of DC, OGP and classic.

pfefferle:
@dshanske I started a MR, to simplify the meta parser: https://github.com/pfefferle/wordpress-webmention/pull/272/files

It is a fallback, so I think it should be as simple as possible and we should not transform the data into to many different formats/arrays.

dshanske:
I have to look at it in detail after work, but it looks much cleaner.

includes/handlers/class-webmention-handler-meta.php
dshanske: Because the og tags are often og:type:article...this sorts them into an array based on the hierarchy.

includes/handlers/class-webmention-handler-meta.php
dshanske: When the properties are parsed out.

includes/handlers/class-webmention-handler-meta.php
dshanske: If you look at the code, it is deliberate as it tries to separate out the hyphenated properties into a property hierarchy....so it is parsing it to an array structure.

includes/handlers/class-webmention-handler-jsonld.php
dshanske: Sorry about that. Was already thinking MF2

includes/handlers/class-webmention-handler-meta.php
```diff
@@ -0,0 +1,89 @@
...
+				continue;
+			}
+
+			$meta[ $meta_name ] = $meta_value;
+		}
+
+		$this->add_properties( $meta );
+
+		// OGP has no concept of anything but mention so it is always a mention.
+		$this->webmention_item->set__response_type = 'mention';
```
pfefferle, line 54: @dshanske shouldn't we used the comment-type class here in some way?

includes/handlers/class-webmention-handler-jsonld.php
```diff
@@ -0,0 +1,109 @@
...
+			$jsonld[] = json_decode( $content, true );
+		}
+		$jsonld = array_filter( $jsonld );
+		if ( 1 === count( $jsonld ) && wp_is_numeric_array( $jsonld[0] ) ) {
+			$jsonld = $jsonld[0];
+		}
+
+		// Set raw data.
+		$this->webmention_item->set__raw( $jsonld );
+		$this->webmention_item->set__response_type = 'mention';
```
pfefferle, line 42: @dshanske shouldn't we use the comment-type class here?

dshanske: For the mf2 one yes, but for the other handlers, it's always mention

dshanske: What I need to do is set up the class mapper function to pull from the comment types

pfefferle:
Thanks! I think we should invest also in some tests, so that refactoring is simpler/easier!

dshanske:
Having a test running the code allowed me to see some issues with it and modify appropriately. The MF2 code that finds the representative code is working properly, the JSON-LD code and Meta Code also seems to work so far.

includes/class-webmention-tools.php
```diff
@@ -53,7 +53,39 @@ public static function register_routes() {
...
 	public static function read( $request ) {
 		$url              = $request->get_param( 'url' );
+		$request = new Webmention_Request();
+		$return  = $request->fetch( $url );
 
+		if ( is_wp_error( $return ) ) {
+			return $return;
+		}
+		$json = array();
+		$meta = new Webmention_Handler_Meta();
```
pfefferle, line 63: can we use this to implement the handler iteration? so that you do not have to initiate all handlers here?

composer.json
```diff
@@ -37,7 +37,7 @@
       "Dealerdirect\\Composer\\Plugin\\Installers\\PHPCodeSniffer\\Plugin::run"
     ],
     "prefix-mf2": [
-			"@php ./vendor/humbug/php-scoper/bin/php-scoper add-prefix --prefix=Webmention --output-dir=./libraries/mf2 --config=config/php-scoper/mf2.inc.php --force --quiet"
+			"@php ./vendor/humbug/php-scoper/bin/php-scoper add-prefix --prefix=Webmention --output-dir=./includes/handlers/libraries/mf2 --config=config/php-scoper/mf2.inc.php --force --quiet"
```
pfefferle, line 40: I would not move it... Yes it is used by a handler, but that does not mean that it has to be in a sub-folder of the handlers... I would keep it in libraries (root folder)...

includes/entities/class-webmention-item.php
```diff
@@ -134,13 +137,16 @@ public function __call( $method, $params ) {
 	}
 
 	public function set( $key, $value, $overwrite = false ) {
+		if ( in_array( $key, array( 'raw', 'site_name', 'response_type' ), true ) ) {
+			$key = '_' . $key;
```
pfefferle, line 141: this might cause errors if you miss adding new params. I would keep the extra underscore, to show the difference even on the setter and getter

includes/handlers/class-webmention-handler-mf2.php
```diff
@@ -0,0 +1,473 @@
...
+					),
+				);
+			}
+		}
+	}
+
+	/**
+	 * Returns the first item in $val if it's a non-empty array, otherwise $val itself.
+	 */
+	protected function first( $val ) {
```
pfefferle, line 377: can we use a more descriptive name here?

includes/entities/class-webmention-item.php
pfefferle: or we introduce something like `set_meta( 'key', 'value' )` to handle the underscores... If we add new params, we shouldn't add them to multiple locations, array, settings...

pfefferle: perhaps we should not prefix the class attributes at all...

pfefferle:
@dshanske the microformats handler is very fragile at the moment, so I would now vote for unit tests. It is hard to test other things if the mf2 handler throws array errors.

includes/handlers/class-webmention-handler-mf2.php
```diff
@@ -437,7 +437,7 @@ protected function get_location( $mf ) {
 	 * @param array Author array.
 	 */
 	protected function get_author( $mf_array ) {
-		$properties = $this->get_property( $mf_array, 'author' );
+		$properties = current( $this->get_property( $mf_array, 'author' ) );
```
pfefferle, line 440: this throws an error if `get_property` does not return an array. try `https://notiz.blog`.

includes/entities/class-webmention-item.php
```diff
@@ -152,6 +152,23 @@ public function set( $key, $value, $overwrite = false ) {
 			$value = webmention_sanitize_html( $value );
 		}
 
+		// Standardize Author Property to h-card
+		if ( 'author' === $key ) {
```
pfefferle, line 144: I would vote for an Author_Item.

includes/handlers/class-webmention-handler-mf2.php
dshanske: Shouldn't it always be an array in mf2 json?

dshanske:
I agree, now that I have done some manual tests, unit tests are next before post type discovery

includes/handlers/class-webmention-handler-mf2.php
pfefferle: maybe, but in this case it isn't. maybe, because `notiz.blog` is an `h-feed` and not an `h-entry`.

dshanske: the representative function is supposed to extract the correct h-entry

dshanske: I see the problem. It's looking for a representative h-entry...but notiz.blog would be an author/feed page. When would an h-card mention another site?

pfefferle: ???

dshanske: You said you used it on notiz.blog. When would that be the source of a webmention?

pfefferle: Maybe never, but we can not predict the usage of other sites, so the parser should be able to handle such cases. Or one step further the parser should never return php errors or warnings, regardless of the site.

pfefferle: The "legacy" wordpress themes for example are using hfeed on every page...

dshanske: Yes, so the code extracts the h-entry whose u-url matches the source url.

dshanske: Then it does author discovery and sets it to the author property if there isn't one.

dshanske: It was meant to give it a 'clean' object to work off of and save.

dshanske:
Added the response type discovery function and conversion to comment data. Most of the Semantic Linkbacks tests were testing response type. I thought about having the class to be used as a property in the comment type class, but it seems like the algorithm needs a distinct order as rsvps have in-reply-to, so it has to be hard-coded to a degree.

I also fixed the improper coding of tag, to separate person and reply as they have different rules

includes/entities/class-webmention-item.php
```diff
@@ -0,0 +1,304 @@
...
+
+	/**
+	 * Returns the representative entry as a comment array.
+	 *
+	 * return array;
+	 */
+	public function to_commentdata_array() {
+		$published_gmt = $this->published->setTimeZone( 'GMT' );
+		$meta          = array(
+			'mf2_category'    => $this->category,
```
pfefferle, line 277: is it a good idea to call them `mf2_*`? I know that you are using it in other plugins, but what about the data we set using JSON-LD or OpenGraph? Shouldn't it be a more generic prefix? Or do we really need a prefix?

includes/handlers/class-webmention-handler-mf2.php
```diff
@@ -0,0 +1,650 @@
...
+
+	/**
+	 * Returns the response type based on parsed microformats.
+	 * Adds some experimental response types.
+	 * @link http://ptd.spec.indieweb.org/#response-algorithm
+	 *
+	 * @param array $mf_array JSON Array of Parsed Microformats.
+	 * @return string Response Type. Default 'mention'.
+	 */
+	public function response_type_discovery( $mf_array ) {
```
pfefferle, line 80: this is a lot of ifs... can we simplify this by iterating through arrays or so? This also does not use the comment type object at all.

pfefferle: I would like to see that the logic is in the comment-type object, so that it is possible to be used in different contexts (JSON-LD, OGP, ...).

pfefferle: Example: you register a comment-type with the pattern to discover it.

dshanske: I originally thought that the comment object would make sense for this, as I commented...however... RSVP is a data property which might need type validation. I tried to fix tagging by making it only work if the person-tag is of the home url...people may tag multiple people in the same post... 

Let me think about how it could be best done, because using the comment object was my original plan. It would probably have to be a function in the comment object that passed through some information.

dshanske: The problem is....with RSVP for example, an RSVP has an in-reply-to property...so if this is not done in a strict order, it would declare itself a reply, not an RSVP.

pfefferle: we can reuqire an order, no problem!

dshanske: I'm still having trouble with this, much as I'd like to do it, because of the complexity of the problem. I think, to address your issues, I could move some of it into the handler class...

includes/handlers/class-webmention-handler-base.php
```diff
@@ -73,4 +73,14 @@ public function set_webmention_item( $webmention_item ) {
...
+
+	/**
+	 * Returns a property from the raw data in the webmention_item.
+	 * Each handler will implement this to transform raw data appropriately.
+	 *
+	 * @param $string $key Property Key. If not provided all properties returned.
+	 *
+	 * @return mixed Return property or false if not found.
+	 */
+	abstract public function get_raw( $key = null );
```
pfefferle, line 85: It is always the same function, so it should be part of the base handler.

Even further, the raw value is part of the `Webmention_Item` so I would recommend to move it to the Item instead!

dshanske: The mf2 version is different, and eventually, I'd like the json-type version to also be a bit different.

pfefferle: But for now they are the same, so we should start generic and if one needs to be different, it can be overwritten in that class.

includes/entities/class-webmention-item.php
dshanske: We still haven't settled this, but even though they come from json-ld or ogp, they are mf2/jf2 property names.

dshanske:
I added in reacji detection for feature parity with Semantic Linkbacks. I tried to add the logic for whether it is complete. I think we can continue to fill in the cracks. What is missing at the moment?

pfefferle:
@dshanske this is getting way to complicated! Let's please focus on Microformats parsing. I removed the emoji handling for now! There are some basic issues we should focus on first:

* The parser uses the source to get the representative item, but it has to use the target.
* The comment-type checker, does not check for the target url, it simply checks if there is a URL that matches one of the types.

I will focus on tests so we will be able to check for those things.

includes/handlers/class-webmention-handler-mf2.php
```diff
@@ -0,0 +1,659 @@
...
+	 * @return array the h- node or false/
+	 */
+	protected function find_representative_item( $mf, $url ) {
+		$items = $this->get_items( $mf );
+		if ( ! is_array( $items ) || empty( $items ) ) {
+			return false;
+		}
+		// Iterate array
+		foreach ( $items as $item ) {
+			if ( $this->urls_match( $url, $this->get_plaintext( $item, 'url' ) ) ) {
```
pfefferle, line 334: This can not work and is way to simple:

* If there is more than one URL it only checks with the first one.
* The URL has not to be in the URL array, one example is the reply context (`u-in-reply-to`)

includes/entities/class-webmention-item.php
```diff
@@ -268,14 +268,14 @@ public function to_commentdata_array() {
...
+			'comment_author_email' => ifset( $this->author['email'] ),
+			'comment_author_url'   => ifset( $this->author['url'] ),
+			'comment_content'      => $this->get_content(),
+			'comment_date'         => $this->published->format( 'Y-m-d H:i:s' ),
+			'comment_date_gmt'     => $published_gmt->format( 'Y-m-d H:i:s' ),
+			'comment_type'         => $this->response_type,
+			'comment_meta'         => array_filter( $meta ),
 		);
 
 		return apply_filters( 'webmention_item_commentdata_array', array_filter( $comment ), $this->get_raw() );
```
dshanske, line 281: Wondering if we should be returning the entire item object in the filter.

dshanske:
I'm getting confused. Shouldn't get_representative item be run on the source_url? It's supposed to be looking for the item that is representative of the retrieved page, not the item that contains the target URL.

pfefferle:
It should look for the item that references the target URL. If there are more than one object, we need the one that links to the target.

pfefferle:
We can also add this check but it is not the representative by only this check!

pfefferle:
And not all URLs have the source url as `u-url` do they? I think I got a lot of errors with the initial version when running the tests.

dshanske:
The goal of the function was to strip everything in but the object that matches the source. Matching the target is verificaton....I hadn't considered that would be in this function. 

Are you thinking we move verification into the handlers, which means that we'd be looking for the URL in an mf2 property in the MF2 parser, which would make sense, as opposed to just a plaintext search.

pfefferle:
But the function checked for the source url with a fallback to the first item. With this code there is a high false rate. Also a text could have more then one link with a response-type, so we have to also check which type is used to link to the target (that's why I also removed the response type code).

Worst case: The source URL has several objects with no reference to the source URL, so we pick the first one. Then we only look for a response type without checking the URL. Then we add a lot of false comments.

I think we can not do it without checking for the target URL. Even for the representative item check. At least if there is more then one object that references the source URL.

dshanske:
> And not all URLs have the source url as `u-url` do they? I think I got a lot of errors with the initial version when running the tests.

Is there an example of a page that has multiple top level h- objects and does not have a url property? If I look back at get_representative_entry in Semantic Linkbacks...it was looking at the target URL to find the representative entity...If I think about it, the entity that mentions the target is the entity we  care about, even if it doesn't represent the page we are retrieving. I think I am thinking about it in a general mf2 way instead of specifically as it applies to webmentions. I think on that basis, I may realign what I'm thinking.

Do we have a Bridgy example to test? Bridgy has a u-url that doesn't match the source url, and we should definitely make sure that there's no problem there.

dshanske:
Thank you for talking this out. I need to make a fix to the Tools item, as it does not have a target URL field.

pfefferle:
You are right in case of the reply context, therefore you need the main item or the representative item of the page.

pfefferle:
There are several brid.gy tests, but perhaps we should have a look if they are still up to date.

dshanske:
With the latest fixes, this will work as expected. Do we want to merge and start tackling the next piece? This PR is getting kind of long.

PR #272, opened 2020-06-30
simplify parser

pfefferle:
I think we do not need a fully fledged meta-parser, so I think we can simplify the code.

### DIFF
--- a/includes/handlers/class-webmention-handler-meta.php
+++ b/includes/handlers/class-webmention-handler-meta.php
@@ -27,6 +27,8 @@ class Webmention_Handler_Meta extends Webmention_Handler_Base {
 		$dom   = clone $request->get_domdocument();
 		$xpath = new DOMXPath( $dom );
 
+		$meta = array();
+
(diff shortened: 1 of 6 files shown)


includes/handlers/class-webmention-handler-meta.php
```diff
@@ -75,150 +37,29 @@ private function parse_meta( $dom ) {
...
-				} else {
-					$return[ $key ] = $value;
-				}
-			}
+	protected function add_property( $key, $value ) {
+		if ( in_array( $key, array( 'url', 'og:url' ), true ) ) {
+			$this->webmention_item->set_url( $value );
+		} elseif ( in_array( $key, array( 'DC.Title', 'og:title' ), true ) ) {
+			$this->webmention_item->set_name( $value );
 		}
```
pfefferle, line 63: this is only exemplary and I should re-add the missing ones

dshanske: Yes, it does look much simpler.

PR #287, opened 2020-08-29
Fix error notices

dshanske:
Found these issues in the error log when testing the local avatar functionality.

### DIFF
--- a/includes/class-webmention-receiver.php
+++ b/includes/class-webmention-receiver.php
@@ -146,13 +146,15 @@ class Webmention_Receiver {
 			'/endpoint',
 			array(
 				array(
-					'methods'  => WP_REST_Server::CREATABLE,
-					'callback' => array( 'Webmention_Receiver', 'post' ),
-					'args'     => self::request_parameters(),
+					'methods'             => WP_REST_Server::CREATABLE,
+					'callback'            => array( 'Webmention_Receiver', 'post' ),
+					'args'                => self::request_parameters(),
+					'permission_callback' => '__return_true',
 				),
 				array(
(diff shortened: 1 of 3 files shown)


dshanske:
I added another fix that was reported. The permission_callback is now required. For a public endpoint, we need it to always return true. https://make.wordpress.org/core/2020/07/22/rest-api-changes-in-wordpress-5-5/

dshanske:
@pfefferle After we merge this, what do we need to release a stable version, even if everything isn't done, so we can get rid of these errors?

pfefferle:
We have to merge it into the 3.x branch! We can't release the master/trunk yet.

dshanske:
Can you merge this into the master/trunk branch and I'll see if we could set up a 4.0 branch to push fixes onto?

includes/class-webmention-receiver.php
```diff
@@ -146,13 +146,15 @@ public static function register_routes() {
 			'/endpoint',
 			array(
 				array(
-					'methods'  => WP_REST_Server::CREATABLE,
-					'callback' => array( 'Webmention_Receiver', 'post' ),
-					'args'     => self::request_parameters(),
+					'methods'             => WP_REST_Server::CREATABLE,
+					'callback'            => array( 'Webmention_Receiver', 'post' ),
+					'args'                => self::request_parameters(),
+					'permission_callback' => '__return_true',
```
pfefferle, line 152: why is this required?

dshanske: https://make.wordpress.org/core/2020/07/22/rest-api-changes-in-wordpress-5-5/

It throws a warning if you don't as of 5.5

PR #297, opened 2020-09-26
Add Avatar Store

dshanske:
This is a reissue of #288 because of the 5.X pushs in the branch.

The only difference is the naming. It was pointed out, as this does not expire, it isn't actually a cache. It is a store, as profile images remain in the store until replaced by a new profile image.

### DIFF
--- a/includes/class-webmention-avatar-handler.php
+++ b/includes/class-webmention-avatar-handler.php
@@ -302,6 +302,11 @@ class Webmention_Avatar_Handler {
 			}
 
 			if ( wp_http_validate_url( $avatar ) ) {
+				$path = 	self::avatar_url_to_filepath( $avatar );
+				// Check to see if filepath exists.
+				if ( $path && ! file_exists( $path ) ) {
+					return $args;
+				}
 				$args['url'] = $avatar;
 			}
 
--- a/includes/class-webmention-avatar-handler.php
+++ b/includes/class-webmention-avatar-handler.php
@@ -16,6 +16,127 @@ class Webmention_Avatar_Handler {
 
 		// All the default gravatars come from Gravatar instead of being generated locally so add a local default
 		add_filter( 'avatar_defaults', array( $cls, 'anonymous_avatar' ) );
+
+		// Store Avatars Locally
+		add_action( 'comment_post', array( $cls, 'store_avatar' ), 20 );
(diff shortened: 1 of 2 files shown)


dshanske:
The issues/questions here...

- It needs some code to migrate old avatars
- Setting to enable disable it (or at least a feature switch using a const).

dshanske:
The final question is the storage of the avatars. They are currently stored under the upload directory /avatars/$host/$user_name.jpg

Occurs to me what is also missing is 
- Check if the file exists and take action if it doesn't
- If the comment/webmention is deleted, the avatar file should be.

pfefferle:
As I commented in the issue about that... It is thought as a cache!

We should try to use it as a cache and keep the social networks like Twitter as exception.

dshanske:
> As I commented in the issue about that... My idea was to implement a **cache**!
> 
> We should try to use it as a cache and keep the social networks like Twitter as exception.

I think the difference is that a cache expires, a store doesn't. Either way, I needed to redo it because of the 5.X/4.X thing.

dshanske:
Terminology aside, the technique you designed works. It stores an image until the next time that the same author URL is the subject of a webmention and then updates it. Which means that the image is valid until a new image replaces it. If that never happens, then the old image will continue to be used. I just added a code check for a non-existent file. It could be enhanced to trigger a refresh of the webmention in some abstract future.

So, how do we move this forward?

dshanske:
Look at https://wordpress.org/plugins/local-gravatars/ - It's new. I like some of the ideas from here we might be able to add.

dshanske:
I was thinking about this. What if, for locally stored items, instead of storing the full url, we did what the author URL does and stored it as %AVATARSTORE%/filename.jpg and before retrieving, did a string replace? That would ensure it wouldn't break on site migration?

dshanske:
@pfefferle I was looking at how other code does this... instead of username it stores by domain then file hash...so it can store earlier iterations.... thinking of trying that. I was concerned about storage size, so I asked @aaronpk how much his store for webmention.io takes up. At 256x256 pixel files, he has 480,000 avatar photos in less than 6gb. So a single WordPress instance shouldn't go anywhere near that.

dshanske:
Came up with an idea. Instead of storing the local URL in meta, we store the original URL. The code hashes it, checks for whether the file exists, if so it returns the local file.

dshanske:
Closed in favor of #314

PR #305, opened 2021-02-17
Fix issue with response not being passed to response code.

dshanske:
https://wordpress.org/support/topic/error-cannot-find-target-link/

The error isn't related to specifically to what was reported, but it reveals a mistake in not passing the response variable optionally to be used.

### DIFF
--- a/includes/class-webmention-request.php
+++ b/includes/class-webmention-request.php
@@ -92,7 +92,7 @@ class Webmention_Request {
 			return $response;
 		}
 
-		$check = $this->check_response_code( wp_remote_retrieve_response_code( $response ) );
+		$check = $this->check_response_code( wp_remote_retrieve_response_code( $response ), $response );
 		if ( is_wp_error( $check ) ) {
 				return $check;
 		}
@@ -127,7 +127,7 @@ class Webmention_Request {
 			return $response;
 		}
 		$this->response_code = wp_remote_retrieve_response_code( $response );
-		$check               = $this->check_response_code( $this->response_code );
+		$check               = $this->check_response_code( $this->response_code, $response );
 		if ( is_wp_error( $check ) ) {
 			return $check;
 		}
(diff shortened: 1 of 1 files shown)


includes/class-webmention-request.php
```diff
@@ -183,10 +183,11 @@ protected function check_content_type( $content_type ) {
 	 * Check if response is a supported content type
 	 *
 	 * @param int $code Status Code.
+	 * @param array $response Response if Present
 	 *
 	 * @return WP_Error|true return an error or that something is supported
 	 */
-	protected function check_response_code( $code ) {
+	protected function check_response_code( $code, $response = null ) {
```
pfefferle, line 190: If `$response` is optional (default `null`) we should also make the output optional (https://github.com/pfefferle/wordpress-webmention/pull/305/files#diff-ee23ad414e45c72a6f2a72d83420c67460c433f4fa2168ee616a5644047930b3R230). Otherwise wie have to require `$response`.

pfefferle: Same for https://github.com/pfefferle/wordpress-webmention/pull/305/files#diff-ee23ad414e45c72a6f2a72d83420c67460c433f4fa2168ee616a5644047930b3R208

pfefferle: Otherwise it has to be `array|WP_Error` as default: https://developer.wordpress.org/reference/functions/wp_remote_retrieve_header/

dshanske: > If `$response` is optional (default `null`) we should also make the output optional (https://github.com/pfefferle/wordpress-webmention/pull/305/files#diff-ee23ad414e45c72a6f2a72d83420c67460c433f4fa2168ee616a5644047930b3R230). Otherwise wie have to require `$response`.

Well, the output is always mandatory. In a case where response is null, wp_remote_retrieve_header returns an empty string, so it would still return the WP_Error object, just without the extra information.

pfefferle (approved):

dshanske:
@pfefferle I think I have to put in a PR to merge these fixes into the 5.X branch also.

PR #314, opened 2021-05-15
Implement Avatar Store in Branch 5.X

dshanske:
The other was based on 4.X. Creating this pull to actually put it into 5.X for various reasons.


dshanske:
I opted to work on this off of Branch 5.X instead of 4.X because in the older code, we stored author url in a meta property. In the new webmention storage philosophy, we store it properly in author URL and stop trying to be backward compatible with pingbacks.

So, it's better to key things to the author URL, whatever it is, as that is what we consider to be their identity.

dshanske:
The code also now removes the gravatar caching code. Instead, it retrieves even the generic default avatars from gravatar and caches them per author url. It also rewrites the gravatar URLs to the default cache size.

dshanske:
@pfefferle The quality check is failing over md5 hashing. I could go to sha256, but it seemed overkill. Thoughts?

includes/class-webmention-avatar-handler.php
```diff
@@ -12,140 +12,224 @@ public static function init() {
...
 	}
 
 	/**
-	 * Function to retrieve Avatar if stored in meta
+	 * Return upload directory url.
 	 *
+	 * @param string $filepath File Path. Optional.
+	 * @return string URL of upload directory.
+	 */
+	public static function upload_directory_url( $filepath = '' ) {
```
pfefferle, line 40: isn't that exactly the same code as in the `upload_directory` function?

dshanske: One should return the URL, the other the filepath. So, for example.

/var/html/wordpress/wp-content/uploads/webmention/avatar/

vs

https://example.com/uploads/webmention/avatar

pfefferle: Ah I see, one uses `baseurl` and the other `basedir`. But both are using the same filter?

pfefferle: because it shares a lot of the same code, shouldn't we add a helper function with two params... the type and the filepath?

dshanske: Good point, I'll adjust.

dshanske: Do you have any thoughts on the md5 vs sha256 issue?

pfefferle: Lets use md5... this is nothing security relevant, so no problem to use it, in my opinion.

dshanske: I'll redo the function. I tested and it does work perfect and migrate any time the comment is updated.

dshanske:
@pfefferle Do we want to try and update and merge this?

pfefferle:
What do you think about decoupling the storing part from the main avatar class, like we did with the handlers? So that it is possible to add new storing classes, like a db store, a CDN store or...?

pfefferle:
Nice! So let's merge the open PRs then and then I will try to change the structure a bit (maybe add namespaces) to prepare different avatar stores?!?

pfefferle:
Shouldn't we get comment Display ready first? To have the current feature set ready and to have the 5.0 ready to launch?

dshanske:
> Shouldn't we get comment Display ready first? To have the current feature set ready and to have the 5.0 ready to launch?

Good point. I'll redo this one later. Just wanted to close out the PRs. But the display does have a higher priority.

dshanske:
@pfefferle I know you are leaning toward an external solution but would still like to pursue this even if it is off by default

PR #319, opened 2022-02-07
added first comment-type discovery implementation

pfefferle:
including (local) tests

### DIFF
--- a/includes/handlers/class-webmention-handler-mf2.php
+++ b/includes/handlers/class-webmention-handler-mf2.php
@@ -40,6 +40,10 @@ class Webmention_Handler_MF2 extends Webmention_Handler_Base {
 
 		$author = $this->get_representative_author( $item, $data );
 
+		// add response type
+		$response_type = $this->get_response_type( $item, $data, $target_url );
+		$this->webmention_item->set_response_type( wp_slash( $response_type ) );
+
 		$this->set_properties( $item );
 		$this->set_property_author( $author );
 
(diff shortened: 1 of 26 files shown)


includes/handlers/class-webmention-handler-mf2.php
```diff
@@ -529,4 +533,200 @@ public function compare_urls( $needle, $haystack, $schemeless = true ) {
...
+	 */
+	protected function get_response_type( $entry, $mf_array, $target ) {
+		$classes = $this->get_class_mapper();
+
+		// check properties for target-url
+		foreach ( $entry['properties'] as $key => $values ) {
+			// check u-* params
+			if ( in_array( $key, array_keys( $classes ), true ) ) {
+				// check RSVP if available
+				if ( 'rsvp' === $key ) {
```
dshanske, line 676: The rsvp:value...do we want to retain the : typing for rsvp? It's the only scenario where the the value of the property dictates the type, I believe.

includes/handlers/class-webmention-handler-mf2.php
```diff
@@ -529,4 +533,200 @@ public function compare_urls( $needle, $haystack, $schemeless = true ) {
...
+					return 'rsvp:' . $this->get_first_array_item( $values );
```
dshanske, line 677: On an rsvp, there should also be an in-reply-to property. We've never checked that before, but should we going forward?

includes/handlers/class-webmention-handler-mf2.php
```diff
@@ -529,4 +533,200 @@ public function compare_urls( $needle, $haystack, $schemeless = true ) {
...
+				}
+
+				// check "normal" links
+				if ( $this->compare_urls( $target, $values ) ) {
+					return $classes[ $key ];
+				}
+
+				// iterate in-reply-tos
+				foreach ( $values as $obj ) {
```
dshanske, line 686: The class has methods for simplifying these checks.

dshanske (changes requested):
I like the design here, from Semantic Linkbacks, but wanted to discuss a few tweaks.

includes/handlers/class-webmention-handler-mf2.php
```diff
@@ -529,4 +533,200 @@ public function compare_urls( $needle, $haystack, $schemeless = true ) {
...
+	 * @param array $entry the represantative entry
+	 * @param array $mf_array the document
+	 *
+	 * @return string the post-type
+	 */
+	protected function get_response_type( $entry, $mf_array, $target ) {
+		$classes = $this->get_class_mapper();
+
+		// check properties for target-url
+		foreach ( $entry['properties'] as $key => $values ) {
```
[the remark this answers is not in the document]
pfefferle: Good point!

includes/handlers/class-webmention-handler-mf2.php
pfefferle: Can you commit a change for that? You can directly add the change to the branch.

includes/handlers/class-webmention-handler-mf2.php
pfefferle: good point! what about adding the sub-type as comment body?

includes/handlers/class-webmention-handler-mf2.php
pfefferle: I would skip that one for the first version, but can you file an issue, so that we are aware of that extension?

includes/handlers/class-webmention-handler-mf2.php
dshanske: Will work on this after work.

includes/handlers/class-webmention-handler-mf2.php
dshanske: I'm still not sure what you meant about this...As in, it goes in content? That would make sense, I suppose. We'd need to do that outside of this function though.

dshanske:
@pfefferle Made some fixes. What do you think?

dshanske (approved):

PR #322, opened 2022-02-11
restructuring

pfefferle:


dshanske (approved):

dshanske:
Added a working handler to the PR. It's a basic one. Needs tweaking. Also, would want to cache the timezone and title for sites for a bit.

pfefferle:
Nice! Can you also add some tests?

PR #324, opened 2022-02-14
Simplify request

pfefferle:


includes/class-request.php
```diff
@@ -208,62 +215,6 @@ protected function check_content_type() {
...
 
-	/**
-	 * Check if response is a supported content type
-	 *
-	 * @param int $code Status Code.
-	 * @param array $response Response if Present.
-	 *
-	 * @return WP_Error|true return an error or that something is supported
-	 */
-	protected function check_response_code() {
```
pfefferle, line 219: I removed this check because it is very fragile. It misses a lot of valid codes, besides `200`.

* https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#2xx_success
* https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#3xx_redirection

pfefferle: Besides that it should be handled by the WordPress request handler.

dshanske (changes requested):

includes/Handler/class-mf2.php
```diff
@@ -29,16 +29,16 @@ class MF2 extends Base {
...
+		$body = $request->get_body();
+		if ( is_wp_error( $body ) ) {
+			return $body;
 		}
 		if ( ! class_exists( '\Webmention\Mf2\Parser' ) ) {
 			require_once plugin_dir_path( __FILE__ ) . '../../libraries/mf2/Mf2/Parser.php';
 		}
 
 		$source_url = $request->get_url();
-		$parser     = new Parser( $dom, $source_url );
```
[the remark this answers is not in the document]
pfefferle: It broke the tests with the don object... will try to fix it, you have a valid point here...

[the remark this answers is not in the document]
pfefferle: seems that it was something else... changed it to dom...

includes/class-request.php
pfefferle: I am so sorry, but I thought the error handling was still part of the `wp_safe_remote_*` methods. So yes we should re-add it 🙄

dshanske:
@pfefferle I use the parser form under tools to do manual testing of the results. It broke in the updates, so fixed it. Then determined why WP wasn't working, fixed that too.

dshanske (approved):

PR #325, opened 2022-02-16
Comment Handling

pfefferle:
Implements the Webmention Comment/Response UI.

### DIFF
--- a/includes/class-comment.php
+++ b/includes/class-comment.php
@@ -24,89 +24,99 @@ class Comment {
 		register_webmention_comment_type(
 			'repost',
 			array(
-				'label'   => __( 'Reply', 'webmention' ),
+				'label'       => __( 'Repost', 'webmention' ),
+				'desctiption' => __( 'A repost on the indieweb is a post that is purely a 100% re-publication of another (typically someone else\'s) post.', 'webmention' ),
 				// translators: %1$s username, %2$s opject format (post, audio, ...), %3$s URL, %4$s domain
-				'excerpt' => __( '%1$s reposted %2$s on <a href="%3$s">%4$s</a>.', 'webmention' ),
+				'excerpt'     => __( '%1$s reposted %2$s on <a href="%3$s">%4$s</a>.', 'webmention' ),
 			)
 		);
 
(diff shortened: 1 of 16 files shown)


includes/class-comment.php
```diff
@@ -0,0 +1,113 @@
...
+	/**
+	 * Register the comment types used by the Webmention plugin
+	 *
+	 * @return void
+	 */
+	public static function register_comment_types() {
+		register_webmention_comment_type(
+			'repost',
+			array(
+				'label'   => __( 'Reply', 'webmention' ),
```
dshanske, line 27: Isn't this repost?

includes/class-comment.php
```diff
@@ -0,0 +1,113 @@
...
+	public static function init() {
+		self::register_comment_types();
+	}
+
+	/**
+	 * Register the comment types used by the Webmention plugin
+	 *
+	 * @return void
+	 */
+	public static function register_comment_types() {
```
dshanske, line 23: Should you register comment? The reason being the idea of having the comment type registry is to be able to execute per type theming, which WordPress itself doesn't have. Like the idea of switching back to comment instead of reply.

dshanske (approved):
Looking forward to expanding the comment type class to do outputs in future.

includes/class-comment.php
pfefferle: Oops!

includes/class-comment.php
pfefferle: Do we really want to change the default output of `comments`, `pingbacks` and `trackbacks`?

dshanske: I was thinking only if protocol is webmention

dshanske:
That duplication alert from SonarCloud is useful in improving code

includes/class-comment.php
pfefferle: the comment marker made issues lately because the core HTML changed a bit and so it broke themes. I would not recommend to overwrite the comment templates using the walker!

pfefferle:
@dshanske can we merge this anyways, because it also fixes the unit tests!?!

PR #326, opened 2022-03-20
Cloud fest

DerHerrFeldmann:


includes/Handler/class-jsonld.php
```diff
@@ -45,8 +45,8 @@ public function parse( Response $response, $target_url ) {
 		}
 
 		// Set raw data.
-		$this->webmention_item->set_raw( $jsonld );
-		$this->webmention_item->set_response_type( 'mention' );
+		$this->webmention_item->add_raw( $jsonld );
```
dshanske, line 48: Why change the name from set to add? Is there some reason?

pfefferle:
Currently every handler that is used overwrites the informations from the handler before, because he 'sets' the new data without checking if they already exist. The `add` functions are calling the `set` functions in the end, but with a check if they are already set.

includes/Handler/class-jsonld.php
pfefferle: Currently every handler that is used overwrites the informations from the handler before, because he 'sets' the new data without checking if they already exist. The `add` functions are calling the `set` functions in the end, but with a check if they are already set.

pfefferle:
@dshanske what do you think?

PR #327, opened 2022-03-27
Introduce Comment Walker

dshanske:
This introduces a basic version of the comment walker. The html5_comment and comment are just modified versions of default with microformats classes added. If you think these shouldn't be here, I could remove and leave only start_el, which is minimum viable.

The real important one is start_el, which now introduces the idea that the callback property could be an array of callback functions by type. This is proof of concept for the start. Next, start_el will get a call to a default handler for webmentions that can be added to the class, namely, if protocol=webmention call these sets of default handlers.


pfefferle:
What is the benefit of overwriting the default comment templates? We just currently had a change in the default Wordpress output and there might be other changes in the future. So we have to keep track of every comment template change in the future.

dshanske:
> What is the benefit of overwriting the default comment templates? We just currently had a change in the default Wordpress output and there might be other changes in the future. So we have to keep track of every comment template change in the future.

So, you think I should go for the less intrusive just start_el? That's fine. With the new enhancement to the callback arguments...we can solve that as needed. Will revert and work on the next piece.

dshanske:
There...will work on the rest of the logic.

pfefferle:
I would fokus on the facepile thing atm.

dshanske:
> I would fokus on the facepile thing atm.

I was going to start by working on the non-facepile versions...the old, "___ liked ____" templating, and the setup for facepiles to work like 'short_ping'.

includes/class-comment-walker.php
```diff
@@ -0,0 +1,131 @@
...
+				return;
+			}
+		}
+
+		if ( 'comment' === $comment->comment_type ) {
+			add_filter( 'comment_text', array( $this, 'filter_comment_text' ), 40, 2 );
+		}
+
+		// Maintain the original pingback and trackback output.
+		if ( ( 'pingback' === $comment->comment_type || 'trackback' === $comment->comment_type ) && $args['short_ping'] ) {
```
pfefferle, line 83: shouldn't this block be also removed?

dshanske: It can't be...because that would prevent it from calling short_ping. That's why I proposed a way to do that without overriding the entire function.

dshanske:
@pfefferle Would like your input on this idea. I think it is better than what we did in Semantic Linkbacks as it allows wp_list_comments to be used for outputting Facepiles.

The new comment walker adds a variety of different new options to wp_list_comments.  I started the barebones, but I need to continue further.

1. The ability to specify callbacks by comment type, but it does fallback to a single one.
1. Two new arguments. `avatar_only`, which is similar to `short_ping`(which I considered overloading for this). If enabled, it will use the `overlay` argument. Avatar only displays the avatar and the associated markup. 
1. My plan for $overlay, which was used in Semantic Linkbacks for the emoji reaction overlay, to allow for the option of either the emoji reaction overlay...which should be done based on it being a reacji type...which we can determine when it is created rather than at display time. But the new functionality would allow the overlaying of an optional symbol for like, etc. if someone wants to show all the different reaction types intermixed. 

One of the long term goals I was hoping for is to allow for different presentation preferences to be implemented. So, this allows each individual webmention type to be a callback, and it allows for a default that would show just the images. 

Before I go farther, as making things look good is not always my strength, wanted some feedback.

webmention.php
```diff
@@ -63,6 +63,9 @@ function init() {
 	require_once dirname( __FILE__ ) . '/includes/class-comment.php';
 	add_action( 'init', array( '\Webmention\Comment', 'init' ) );
 
+	require_once dirname( __FILE__ ) . '/includes/class-comment-walker.php';
```
pfefferle, line 66: we should add an option/const/add_theme_support to disable the "comment representation" feature of the Webmention plugin.

dshanske: I don't think it is necessary at the place you identify. As of now, the Comment Walker just adds additional options to wp_list_comments. The default is supposed to be false...noticed I committed it as true by accident. 

The options for how to display differently will likely be changing the default display to exclude the mention types, and allowing them to be called separately, either by using the hook we use in Semantic Linkbacks, or the theme calling wp_list_comments for them specifically.

includes/class-comment-walker.php
```diff
@@ -0,0 +1,173 @@
...
+	 * @param array      $args    An array of arguments.
+	*/
+	protected function avatar_only( $comment, $depth, $args ) {
+		$tag = ( 'div' === $args['style'] ) ? 'div' : 'li';
+
+		$title = get_comment_text( $comment, $args );
+		// Optionally overlay an icon.
+		$overlay = '';
+		?>
+		<<?php echo $tag; ?> id="comment-<?php comment_ID(); ?>" <?php comment_class( '', $comment ); ?>>
```
pfefferle, line 161: use a template for this?

pfefferle:
I am not completely sure I understand everything you mentioned, but I would love to see the restult. Let's get this ready, because the SL plugin is falling apart and if I try to fix things, it breaks something else!

dshanske:
I have been working on it. Haven't committed everything

pfefferle:
I am curious 😊

includes/class-comment-walker.php
dshanske: The new code allows someone to write callbacks, the way comments are currently done, per comment type. The goal was to try and make it as close to the stock method as possible. I can move this to a template, but it shouldn't be able to be overwritten here as it is further upstream.

dshanske:
Okay, a few more things to do, doing this in an integrated way instead of Semantic Linkbacks, which used a completely different function, took some thinking. So, if you call wp_list_comments with the avatar-only argument set, the walker code will call a new function which displays only the avatar. It adds a simple style sheet which displays it in a row.

The new push introduces an optional emoji per comment type which will be overlaid over the image. So, you can disable that, and do a row of each type(and still call the new function to get the emoji for that type), or a row of every type with the icons interspersed.

That makes the next step removing registered types from the default list displayed, and adding it where the comment form is, the way Semantic Linkbacks does, and implementing the settings.

pfefferle:
Should we add a default "view" to the register comment type?

dshanske:
Well, my assumption was that the default view for everything registered would be the avatar only view. But what do you think?

dshanske:
@pfefferle I'd like to merge this in. By itself, it only adds the functionality for display. I think it would be better to add the settings for display and the hooks/templates to show it in a separate PR.

To go over what this does, as you claimed there was some confusion.

- It introduces icons to each registered comment type that can be used for display purposes. They are based on indieweb.org/posts
- It introduces a comment walker that is a small override of the stock one. However, the comment walker is not the same as the more aggressive one I put into Semantic Linkbacks.
- The comment walker allows for processing extra arguments to wp_list_comments. The good thing about the approach is that if someone introduces an alternative comment walker, or a comment callback function, it will override this neatly.
- The callback argument in the default comment walker usually accepts a single function name. One of the modifications is it allows for that, or an array, allowing for different callbacks for different comment type.

So, there is no need for templating or otherwise, because this works entirely within how WordPress already works.

includes/class-comment-walker.php
```diff
@@ -0,0 +1,172 @@
...
+	 *
+	 * @see Walker::end_el()
+	 * @see wp_list_comments()
+	 *
+	 * @param string     $output      Used to append additional content. Passed by reference.
+	 * @param WP_Comment $data_object Comment data object.
+	 * @param int        $depth       Optional. Depth of the current comment. Default 0.
+	 * @param array      $args        Optional. An array of arguments. Default empty array.
+	 */
+	public function end_el( &$output, $data_object, $depth = 0, $args = array() ) {
```
pfefferle, line 124: I think we can remove this, because it has no changes at all.

includes/class-comment-walker.php
```diff
@@ -0,0 +1,172 @@
...
+		$tag = ( 'div' === $args['style'] ) ? 'div' : 'li';
+
+		$title = get_comment_text( $comment, $args );
+		// Optionally overlay an icon.
+		$overlay = '';
+		if ( $args['overlay'] ) {
+			$overlay = '<span class="emoji-overlay">' . get_webmention_comment_icon( $comment ) . '</span>';
+		}
+		?>
+		<<?php echo $tag; ?> id="comment-<?php comment_ID(); ?>" <?php comment_class( array( 'u-comment', 'h-cite', 'avatar-only' ), $comment ); ?>>
```
pfefferle, line 160: shouldn't we use a template here?

includes/class-comment-type.php
```diff
@@ -25,14 +25,20 @@ final class Comment_Type {
...
 	 * @var string $singular
 	 */
 	public $singular;
 
+	/**
+	 * Single Character Emoji Representation of the Comment Type. Optional.
+	 *
+	 * @var string $icon
+	 */
+	public $icon;
```
pfefferle, line 40: what about adding dashicons instead? https://developer.wordpress.org/resource/dashicons/

includes/class-comment.php
```diff
@@ -25,7 +25,8 @@ public static function register_comment_types() {
 			'repost',
 			array(
 				'label'       => __( 'Repost', 'webmention' ),
-				'desctiption' => __( 'A repost on the indieweb is a post that is purely a 100% re-publication of another (typically someone else\'s) post.', 'webmention' ),
+				'description' => __( 'A repost on the indieweb is a post that is purely a 100% re-publication of another (typically someone else\'s) post.', 'webmention' ),
+				'icon' => '♺',
```
pfefferle, line 29: https://developer.wordpress.org/resource/dashicons/#controls-repeat

includes/class-comment.php
```diff
@@ -36,6 +37,7 @@ public static function register_comment_types() {
 			array(
 				'label'       => __( 'Like', 'webmention' ),
 				'description' => __( 'Like', 'webmention' ),
+				'icon' => '👍',
```
pfefferle, line 40: https://developer.wordpress.org/resource/dashicons/#heart

includes/class-comment.php
```diff
@@ -46,6 +48,7 @@ public static function register_comment_types() {
 			array(
 				'label'       => __( 'Favorite', 'webmention' ),
 				'description' => __( 'A favorite is a common webaction on many silos (like Flickr, Twitter), typically visually indicated with a star symbol that fills in with a color when activated (pink, orange).', 'webmention' ),
+				'icon' => '⭐',
```
pfefferle, line 51: https://developer.wordpress.org/resource/dashicons/#star-filled

includes/class-comment.php
```diff
@@ -56,6 +59,7 @@ public static function register_comment_types() {
 			array(
 				'label'       => __( 'Tag', 'webmention' ),
 				'description' => __( 'Tags or tagging refers to categorizing or labeling content, your own or others (tag-reply), with words, phrases, names, or other information, optionally linked to specific people, events, locations, such as the practice of tagging posts being about certain people (person-tag), like tagging people or other items where (area-tag) they\'re depicted in a photo.', 'webmention' ),
+				'icon' => '📌',
```
pfefferle, line 62: https://developer.wordpress.org/resource/dashicons/#tag

includes/class-comment.php
```diff
@@ -66,6 +70,7 @@ public static function register_comment_types() {
 			array(
 				'label'       => __( 'Bookmark', 'webmention' ),
 				'description' => __( 'A bookmark (or linkblog) is a post that is primarily comprised of a URL, often title text from that URL, sometimes optional text describing, tagging, or quoting from its contents.', 'webmention' ),
+				'icon' => '🔖',
```
pfefferle, line 73: https://developer.wordpress.org/resource/dashicons/#book

includes/class-comment.php
```diff
@@ -76,6 +81,7 @@ public static function register_comment_types() {
 			array(
 				'label'       => __( 'Listen', 'webmention' ),
 				'description' => __( 'A "listen" is a passive type of post used to publish a song (music or audio track, including concert recordings or DJ sets) or podcast that you have listened to.', 'webmention' ),
+				'icon' => '🎧',
```
pfefferle, line 84: https://developer.wordpress.org/resource/dashicons/#format-audiohttps://developer.wordpress.org/resource/dashicons/#format-audio

includes/class-comment.php
```diff
@@ -86,6 +92,7 @@ public static function register_comment_types() {
 			array(
 				'label'       => __( 'Watch', 'webmention' ),
 				'description' => __( 'A watch is a semi-passive type of post used to publish that you have watched a video (movie, TV, film), or a live show (theater, concert).', 'webmention' ),
+				'icon' => '📺',
```
pfefferle, line 95: https://developer.wordpress.org/resource/dashicons/#video-alt3

includes/class-comment.php
```diff
@@ -95,6 +102,7 @@ public static function register_comment_types() {
 			'read',
 			array(
 				'label'       => __( 'Read', 'webmention' ),
+				'icon' => '📖',
```
pfefferle, line 105: https://developer.wordpress.org/resource/dashicons/#book

includes/class-comment.php
```diff
@@ -106,6 +114,7 @@ public static function register_comment_types() {
 			array(
 				'label'       => __( 'Follow', 'webmention' ),
 				'description' => __( 'Follow is a common feature (and often UI button) in silo UIs (like Twitter) that adds updates from that profile (typically a person) to the stream shown in an integrated reader, and sometimes creates a follow post either in the follower\'s stream ("… followed …" or "… is following …") thus visible to their followers, and/or in the notifications of the user being followed ("… followed you").', 'webmention' ),
+				'icon' => '👣',
```
pfefferle, line 117: https://developer.wordpress.org/resource/dashicons/#admin-users

includes/class-comment.php
pfefferle: maybe use https://developer.wordpress.org/resource/dashicons/#tag instead because of "read" activity

includes/class-comment.php
pfefferle: maybe use https://developer.wordpress.org/resource/dashicons/#admin-post instead because of "bookmark" activity

pfefferle:
Sounds good to me, but I still had some small (inline)questions.

includes/class-comment-type.php
dshanske: Somewhat limited in their expandability though.

pfefferle: That’s true.

dshanske: Maybe better off just adding a filter and people can change it

webmention.php
```diff
@@ -40,9 +40,10 @@
  * Plugin Version Number used for caching.
  */
 function version() {
-	return '5.X';
-}
+	$meta = get_plugin_meta( false );
```
dshanske, line 43: Where does this get_plugin_meta function come from? It isn't in core. And it causes the code to crash

dshanske: Fixed it myself.

pfefferle: It is defined at the end of this file. it calles the core function `get_plugin_data` but it seems to be only loaded in the wp-admin context.

dshanske:
@pfefferle How's that?

webmention.php
dshanske: I fixed it, using the function that it itself calls, which is not only in an admin context

dshanske:
What else is pending before merge?

pfefferle (approved):
LGTM

PR #335, opened 2022-06-07
Introduce Webmention Comments Template

dshanske:
This introduces a simple webmentions comment template. It actually does it differently than the Semantic Linkbacks template. Instead of using the comment form hooks, it actually loads the template, if the setting is enabled, then that template loads the theme template. So the original comment template, which is filtered to only show non-webmention comments, shows as intended by the theme author. And the new template just loads before it.

This is a very barebones implementation, and I expect we'll iterate on it with more features.


dshanske:
@pfefferle Can you have a look?

pfefferle:
Will do! I had vacations until yesterday ;)

dshanske:
Didn't realize..  hope you had fun

dshanske:
> Will do! I had vacations until yesterday ;)

@pfefferle Just waiting on this...not trying to be a bother.

pfefferle:
I get an error when I visit the post permalink site:

```PHP
Warning: require_once(/var/www/html/wp-content/themes/twentytwentytwo/comments.php): failed to open stream: No such file or directory in /var/www/html/wp-includes/template.php on line 770 Fatal error: require_once(): Failed opening required '/var/www/html/wp-content/themes/twentytwentytwo/comments.php' (include_path='.:/usr/local/lib/php') in /var/www/html/wp-includes/template.php on line 770 
```

dshanske:
How does it look overall? I see you fixed a locate issue

dshanske:
@pfefferle Any chance on this one?

pfefferle:
I am working on some dummy/test data, otherwise it is hard to see the result.

pfefferle (approved):
LGTM

PR #339, opened 2022-07-16
Update Dependencies, Actions and WP Requestor

dshanske:
I updated the dependencies on this, and ran the tests, but noticed an issue with the wp rest api call...I found a way to reduce a function call by asking for the authorship data to be embedded. This will change the creation of future tests.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -8,9 +8,9 @@ jobs:
     name: New tag
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@master
+    - uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


pfefferle (commented):
I am not sure I understand the embed thing

dshanske:
> I am not sure I understand the embed thing

https://developer.wordpress.org/rest-api/using-the-rest-api/linking-and-embedding/ - It just saves you from having to poll for the author, it adds the author object to the return. I didn't notice it before.

pfefferle:
Awesome!

pfefferle (approved):
LGTM!

PR #345, opened 2023-01-04
Comment form template: add hooks before and after the form

jeherve:
Would you be open to adding hooks in the different templates available in the plugin?  This would allow one to add content before or after the webmention form, without having to overwrite the full template.

For example, one could add an HTML element between the comment form and the webmention form, without having to rely on custom CSS or having to define their own custom `templates/webmention-comment-form.php` via the existing `webmention_comment_form` filter.

```php
add_action(
	'webmention_comment_form_template_before',
	function () {
		echo '<hr class="styled-separator is-style-wide" aria-hidden="true">';
	}
);
```

What do you think?

### DIFF
--- a/templates/webmention-comment-form.php
+++ b/templates/webmention-comment-form.php
@@ -1,3 +1,9 @@
+<?php
+/**
+ * Hook to add custom content before the webmention form added to the comment form.
+ */
+do_action( 'webmention_comment_form_template_start' );
+?>
 <form id="webmention-form" action="<?php echo get_webmention_endpoint(); ?>" method="post">
 	<p>
 		<label for="webmention-source"><?php echo get_webmention_form_text( get_the_ID() ); ?></label>
@@ -11,3 +17,9 @@
 	<input id="webmention-format" type="hidden" name="format" value="html" />
 	<input id="webmention-target" type="hidden" name="target" value="<?php the_permalink(); ?>" />
 </form>
+<?php
+/**
+ * Hook to add custom content after the webmention form added to the comment form.
+ */
+do_action( 'webmention_comment_form_template_end' );
+?>
--- a/templates/webmention-comment-form.php
+++ b/templates/webmention-comment-form.php
@@ -2,7 +2,7 @@
 /**
  * Hook to add custom content before the webmention form added to the comment form.
  */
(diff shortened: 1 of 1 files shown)


templates/webmention-comment-form.php
```diff
@@ -1,13 +1,31 @@
...
 	</p>
 	<p>
-		<input id="webmention-source" type="url" autocomplete="url" name="source" placeholder="<?php esc_attr_e( 'URL/Permalink of your article', 'webmention' ); ?>" />
+		<input
+			id="webmention-source"
+			type="url"
+			autocomplete="url"
+			name="source"
+			placeholder="<?php esc_attr_e( 'URL/Permalink of your article', 'webmention' ); ?>"
+		/>
```
jeherve, line 18: I only updated this to make sonarcloud happier (it was reporting a fail because the line was too long).

pfefferle: I have to remove sonarcube, it is way to pedantic :)

pfefferle: please ignore it by now, I will switch to PHPCS!

pfefferle:
Thanks for you PR, can you create the same PR for the 5.X branch please?

templates/webmention-comment-form.php
```diff
@@ -1,13 +1,31 @@
+<?php
+/**
+ * Hook to add custom content before the webmention form added to the comment form.
+ */
+do_action( 'webmention_comment_form_template_start' );
```
pfefferle, line 5: isn't `before`/`after` or `pre`/`post` the WordPress way to go?

jeherve:
> can you create the same PR for the 5.X branch please?

Sure thing. I've updated this PR to be against `5.X`.

templates/webmention-comment-form.php
jeherve: 👍 I've removed that clean-up commit.

templates/webmention-comment-form.php
jeherve: I'm happy to go with `before`/`after`. Done in c10bcaf86cc1e560c2606dcf44ac0721814a1304

pfefferle (approved):
LGTM, thanks!

PR #348, opened 2023-01-06
DB migration

pfefferle:
To migrate:

- [x] `semantic_linkbacks_source` -> `webmention_source`
- [x] `semantic_linkbacks_avatar` -> `avatar`
- [x] `semantic_linkbacks_type` -> `protocol`
- [x] `semantic_linkbacks_canonical` -> `webmention_canonical`
- [x] `semantic_linkbacks_author_url` -> `webmention_author_url`
- [x] added `protocol` 

Process:

- [x] Call migration on update procedure.

### DIFF
--- a/includes/class-db.php
+++ b/includes/class-db.php
@@ -8,10 +8,10 @@ class DB {
 	 *
 	 * @var int
 	 */
-	private static $target_version = 1.0;
+	private static $target_version = '1.0.0';
 
 	private static function get_target_version() {
-		return self::$db_version;
+		return self::$target_version;
 	}
 
 	private static function get_version() {
@@ -26,11 +26,11 @@ class DB {
 	private static function is_latest_version() {
 		$current_version = self::get_version();
 
-		return $current_version === self::$db_version;
+		return (bool) version_compare( $current_version, self::$target_version, '==' );
 	}
 
 	/**
-	 * Updates the database structure if necessary
+	 * Updates the database structure if necessary.
 	 */
 	public static function update_database() {
 		if ( self::is_latest_version() ) {
@@ -38,11 +38,14 @@ class DB {
 		}
 
 		$version_from_db = self::get_version();
-		if ( $version_from_db < 1.0 ) {
+		if ( version_compare( $version_from_db, '1.0.0', '<' ) ) {
 			global $wpdb;
 
 			// 1. rename comment meta
 			self::update_commentmeta_key( 'semantic_linkbacks_avatar', 'avatar' );
+			self::update_commentmeta_key( 'semantic_linkbacks_author_url', 'webmention_author_url' );
+			self::update_commentmeta_key( 'semantic_linkbacks_canonical', 'webmention_canonical' );
+			self::update_commentmeta_key( 'semantic_linkbacks_source', 'webmention_source' );
 			// 2. migrate comment type
 		}
 
@@ -50,7 +53,7 @@ class DB {
 	}
 
 	/**
-	 * Rename meta keys
+	 * Rename meta keys.
 	 *
 	 * @param string $old The old commentmeta key
 	 * @param string $new The new commentmeta key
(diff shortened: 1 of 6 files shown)


includes/class-db.php
```diff
@@ -0,0 +1,193 @@
...
+	 * Migrate webmentions to comment types
+	 *
+	 */
+	public static function update_comment_type() {
+		global $wpdb;
+		foreach ( array( 'mention', 'reply', 'repost', 'like', 'favorite', 'tag', 'bookmark', 'invited', 'listen', 'watch', 'read', 'follow' ) as $type ) {
+			$comment_id_list = get_comments(
+				array(
+					'type'       => 'webmention',
+					'meta_key'   => 'semantic_linkbacks_type',
```
pfefferle, line 108: @dshanske shouldn't it be easier to search for

```
'meta_query' => array(
					array(
						'key'     => 'semantic_linkbacks_type',
						'compare' => 'EXISTS',
					),
				),
```

The current list might miss some posts like RSVP vor example.

includes/class-db.php
```diff
@@ -0,0 +1,193 @@
...
+				$old
+			)
+		);
+	}
+
+	/**
+	 * Migrate webmentions to comment types
+	 *
+	 */
+	public static function update_comment_type() {
```
pfefferle, line 102: We have to delete `semantic_linkbacks_type` at the end!?!

pfefferle:
@dshanske what do you think?

pfefferle:
@gregorlove you also said, that you would love to help... maybe you can also have a look...

dshanske (changes requested):
What about migrating the author URL to the author_url property in the comment? The new version uses the one in the comment data structure? Also, there is no canonical URL field in the new code. It is just storing it in URL.

includes/class-db.php
```diff
@@ -94,20 +94,23 @@ public static function update_options_key( $old, $new ) {
 	public static function migrate_to_1_0_0() {
 		// 1. rename comment meta
 		self::update_commentmeta_key( 'semantic_linkbacks_avatar', 'avatar' );
-		self::update_commentmeta_key( 'semantic_linkbacks_author_url', 'webmention_author_url' );
-		self::update_commentmeta_key( 'semantic_linkbacks_canonical', 'webmention_canonical_url' );
```
dshanske: This should be self::update_commentmeta_key( 'semantic_linkbacks_canonical', 'url' );

includes/class-db.php
```diff
@@ -94,20 +94,23 @@ public static function update_options_key( $old, $new ) {
 	public static function migrate_to_1_0_0() {
 		// 1. rename comment meta
 		self::update_commentmeta_key( 'semantic_linkbacks_avatar', 'avatar' );
-		self::update_commentmeta_key( 'semantic_linkbacks_author_url', 'webmention_author_url' );
```
dshanske: This should migrate to $comment->comment_author_url

pfefferle: I had a look at the code and this is mostly already done. Should I do it anyways?

includes/class-db.php
pfefferle: done

includes/class-db.php
pfefferle: @dshanske `comment_author_url` currently has the brid.gy and `semantic_linkbacks_author_url` is the real URL. I think we have to re-check the code here to get that right!

dshanske (approved):

includes/class-db.php
dshanske: > @dshanske `comment_author_url` currently has the brid.gy and `semantic_linkbacks_author_url` the real URL. I think we have to re-check the code here to get that right!

https://github.com/pfefferle/wordpress-webmention/blob/5.X/includes/Entity/class-item.php#L388 - You can see right here 

https://github.com/pfefferle/wordpress-webmention/blob/5.X/includes/Entity/class-item.php#L383 - And here

So, the author_url will only be the Bridgy URL if it can't find anything else, as a fallback.

pfefferle: And where do we store the bridgy URL?

dshanske: That is the source_url

dshanske: It is still registered as webmention_source_url

PR #349, opened 2023-01-22
Fixes an Issue in Comment Types

dshanske:
Started doing some user testing on a live test site and noticed this bug with the comment types. We never set the singular property and that is used in the Comment Admin screen.


pfefferle (approved):
👍

PR #354, opened 2023-03-02
Use available `$vouch` instead of `$params['vouch']`

jeremyfelt:
This would result in a couple PHP warnings and cause the `webmention_vouch_url` to always store as an empty string, even if a vouch URL was provided.

```bash
wp> $vouch = urldecode( $params['vouch'] );
=> string(0) ""
wp> esc_url_raw( $vouch );
=> string(0) ""
```


pfefferle (approved):
👍

pfefferle (approved):
👍

PR #358, opened 2023-03-19
Check version when settings page is loaded

dshanske:
This is the simplest way of ensuring migration is checked after activation. It runs the upgrade check each time you visit the settings page, so that avoids checking it on every page load, in the event it doesn't trigger on upgrade.

I thought about more complicated options, but this covers us in most cases.

### DIFF
--- a/includes/class-admin.php
+++ b/includes/class-admin.php
@@ -230,6 +230,10 @@ class Admin {
 	 * Load settings page
 	 */
 	public static function settings_page() {
+		require_once dirname( __FILE__ ) . '/class-db.php';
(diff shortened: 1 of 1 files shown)


pfefferle:
@dshanske is this enough? Is there a possibility that something breaks if someone do not visit the settings site after upgrading?

dshanske:
It's a backup for the activation hook

dshanske:
Do you want to consider merging this and I'll see about other triggers as well?

pfefferle:
@dshanske sorry, I was at the CloudFest Hackathon and forgot to approve the PR before!

PR #362, opened 2023-04-05
Added missing space

edent:


pfefferle (changes requested):
You have to change this in the code too: https://github.com/pfefferle/wordpress-webmention/blob/d4bb5fb09f4622ea8d3e705862ab115b30b359ac/templates/webmention-settings.php#L34

pfefferle (approved):
LGTM, thanks!

PR #364, opened 2023-04-05
Templating

chrisbergr:
Added functionality for overwriteable templates


pfefferle (commented):
Oops, I was in the wrong PR, ignore my "Approve", I want to wait, what @dshanske thinks about it!

dshanske:
I spent a lot of time thinking about this.

In the current version of the code, there is a way to do per-comment type callbacks. It changes the signature of the callback argument in https://developer.wordpress.org/reference/functions/wp_list_comments/ to allow for an array of callback functions. So, that allows for individual templating as well as fallback with the original way of taking over a callback entirely.

The new code also allows for two new arguments, 'avatar' for avatar only, and 'overlay' to overlay an emoji symbol on top of the avatar. The reasoning for this is 100% backward compatibility. It means that anyone using these classic methods would be able to do what they always did, and any new person take advantage of this progressive enhancement.

In this PR, in the avatar generated version of this, the direct markup is replaced with a call to a new webmention template function.  The same function is also used in a new html5_comment(but not the non-html5 comment version). But that means it is handling local comments as well, which may be out of scope for this plugin. I'd like it to render all comments with mf2 markup by default, but we're back to...is that within scope?

There is also an issue with passing arguments through via locate_template. We note a compatibility with WP4.9 and above. The Argument option wasn't introduced till WP5.5.

With the templates, you have reaction-item and comment-item. However, if we are keeping with the usual way templates work(https://developer.wordpress.org/themes/basics/template-hierarchy), the last part should probably be the comment type. So, if reaction is the generic template, and you want different ones for different comment types, it might be reaction-like or comment-like...  But a file called comment.php for the generic might be confusing...

dshanske:
I think we might want to scale it down slightly, and then merge, and we can decide if we want to build it up a bit more.

dshanske:
I was thinking we need the following basic templates that can be overridden in the theme. The avatar only reaction and comment style rendering that has to take into account reactions rendered as comments, bringing in the textual description version

chrisbergr:
> There is also an issue with passing arguments through via locate_template. We note a compatibility with WP4.9 and above. The Argument option wasn't introduced till WP5.5.

You're right, I did not consider supporting a 6 year old version. I missed the compatibility note.

> With the templates, you have reaction-item and comment-item. However, if we are keeping with the usual way templates work(https://developer.wordpress.org/themes/basics/template-hierarchy), the last part should probably be the comment type. So, if reaction is the generic template, and you want different ones for different comment types, it might be reaction-like or comment-like... But a file called comment.php for the generic might be confusing...

I have thought very carefully about the template hierarchy. The last part is in fact the comment type. You can use reaction-item.php to define the output for all reaction items, and use reaction-item-like.php to define only the like list items. Ive done this in my test by overwriting the Avatar size only for likes https://cho.bz/t/mA

PR #367, opened 2023-04-06
Check if 404 template exists

### LINKED ISSUE #366: Not all themes have 404 templates
I'm seeing lots of these errors:

`PHP Fatal error:  Uncaught ValueError: Path cannot be empty in /home/blog/wp-content/plugins/webmention/templates/webmention-comment.php:10`

https://github.com/pfefferle/wordpress-webmention/blob/216b1b9727437a3c3d44d51476e4c551a14a9f6e/templates/webmention-comment.php#L10

That should probably be replaced with:

```php
if ( null !== get_query_template( '404' ) ) {
   include( get_query_template( '404' ) );
}
```

Which is similar to the approach in https://github.com/pfefferle/wordpress-webmention/blob/main/includes/class-http-gone.php

edent:
Fixes #366


templates/webmention-comment.php
```diff
@@ -7,7 +7,12 @@
 if ( ! $comment ) {
 	status_header( 404 );
 	nocache_headers();
-	include get_query_template( '404' );
+	// check if theme has a 404.php template
+	$template_404 = get_query_template( 404 );
+	// return 404 template
+	if ( $template_404 ) {
+		include get_query_template( '404' );
```
pfefferle, line 14: I think we could re-use `$template_404` here, instead of querying it again?

pfefferle:
Will that result in an empty page (White screen) if there is no template?

edent:
Yes - which is no different from what happens now.

pfefferle:
sure, but if we are already improving it... ;)

templates/webmention-comment.php
```diff
@@ -7,7 +7,12 @@
 if ( ! $comment ) {
 	status_header( 404 );
 	nocache_headers();
-	include get_query_template( '404' );
+	// check if theme has a 404.php template
+	$template_404 = get_query_template( 404 );
+	// return 404 template
+	if ( $template_404 ) {
+		include $template_404; 
+	}
```
pfefferle, line 15: what do you think about:

```PHP
	if ( $template_404 ) {
		include $template_404;
		die();
	} else {
		wp_die(
			__( 'Page not found', 'webmention' ),
			__( '404', 'webmention' )
		);
	}
```

edent: I think that makes sense.

dshanske:
Actually, the template hierarchy says that if there is no 404 page, it should return index.php.

pfefferle:
Have I used the wrong method @dshanske ?

dshanske:
> Have I used the wrong method @dshanske ?

I'm just asking, shouldn't it redirect to the index instead of dying as a white screen?

dshanske (approved):
Looks good.

Issue #373, opened 2023-04-09
Add comment type to edit comment screen


dshanske:
The comment screen for editing does not show the comment type or other metadata by default. Create a better presentation here and explore bringing back converting one type into another.

edent:
On my comment edit screen, I see this:

![Screenshot showing the metadata](https://user-images.githubusercontent.com/837136/231211025-fd66f7c6-e6a9-4d6f-835c-2730ce539592.png)

I can't edit it - but it does seem to show the metadata.

dshanske:
I want to update that for the new era

dshanske:
Resolved via #375

PR #375, opened 2023-04-11
Improve webmention comment metabox

dshanske:
Attempt to address #373 .

It registers the missing url and avatar meta keys.

URL is supposed to represent the canonical URL.


pfefferle (approved):
LGTM

Issue #346, opened 2023-01-05
Webmentions on Custom Post Types


dshanske:
I found a reason why Custom Post Types webmention sending wouldn't work.

Essentially, we trigger webmentions using do_pings, which is triggered by _publish_post_hook, which is triggered by publish_post.

Since that only applies to the 'post' post type, it means any pending webmentions won't be sent for any other post type until the next time someone publishes a 'post' post type.

We can fix this by creating our own trigger for publish_{post_type} that executes

	if ( ! wp_next_scheduled( 'do_pings' ) ) {
		wp_schedule_single_event( time(), 'do_pings' );
	}

It probably also explains a few more issues.

pfefferle:
Oh, good point!

tmichellemoore:
Thank you @dshanske for diagnosing this custom post type issue. I will not give up on them now!

(Originally published at: https://tmichellemoore.com/?p=151626)

pfefferle:
@dshanske add your tweak for now and think about a different approach for the long term (to not slow down 5.X any more)?

dshanske:
I think I'll do up a PR for both... I don't think it's a problem

dshanske:
Resolved via #376

PR #376, opened 2023-04-11
Add support for CPTs

dshanske:
Addresses #346. There is no concern about extra performance load because it only triggers on the case of a webmention enabled post type, and it checks to see if do_pings is already scheduled.


pfefferle (approved):
LGTM

Issue #372, opened 2023-04-08
Make likes, repost and mention overlay emojis optional


danielcharrua:
Would be great to let users select to turn off/on overlay emojis on likes, repost, mentions, etc...

They are a good way to identify the mentions but do not add any value on design and I think they add "noise" (IMHO) to the mentions.

Thank you.

dshanske:
In my original version, had planned to add options to have a row of images and a single icon at the beginning, the overlay option, etc. We're looking at templating, but some options that don't require code are probably warranted.

dshanske:
Opted to go for this by filter instructions, rather than a setting.

dshanske:
This technically makes this resolved.

PR #377, opened 2023-04-12
Improve Display

dshanske:
This addresses a variety of display issues, specifically.

- #372 Adding an option to disable the overlay
- Fixes a microformats issue as the class is supposed to be the display class, but it is using the parsing class
- Introduces a function to get the canonical URL of a webmention comment
- Introduces a version of the comment rendering from Semantic Linkbacks. The difference between this and the stock default is that is is microformats 2 compliant and adds the citation that was present in that plugin, as well as introducing a hook to add arbitrary metadata
- Introduce hooks proposed in templating PR by @chrisbergr 

This sets up us moving into adding in #364.


includes/class-comment.php
```diff
@@ -70,7 +70,7 @@ public static function register_comment_types() {
 				'singular'    => __( 'Repost', 'webmention' ),
 				'description' => __( 'A repost on the indieweb is a post that is purely a 100% re-publication of another (typically someone else\'s) post.', 'webmention' ),
 				'icon'        => '♻️',
-				'class'       => 'u-repost-of',
+				'class'       => 'p-repost',
```
pfefferle, line 73: wasn't `*-of` or `*-to` the preferred way? https://indieweb.org/reply#How_To

pfefferle:
Shouldn't we work on templating instead of adding so much features/settings in the "old world"? I think it would make all the changes in this PR obsolet.

/cc @chrisbergr

pfefferle:
Why I am not a big fan of overwriting the `html5_comment` function:

In the current PR we only replace the template for Webmentions (which seems not 100% compatible to that in core), which means Webmentions and Comments look slightly different:

Bildschirmfoto 2023-04-12 um 10 07 40

If WordPress changes the default behaviour and/or the CSS changes, it might break Webmention representation.

Most of the themes do not rely on the comment walker template at all and implement their own representation using the comment callback. See the autonomie theme for example: `wp_list_comments( array( 'callback' => 'autonomie_comment', 'format' => '' ) );`.

dshanske:
It was designed to accommodate themes that don't use the comment callback. If you use it, that would take precedence. The same issue would be the case with @chrisbergr 's pr. Which can be implemented by moving this code to template files.

includes/class-comment.php
dshanske: For marking it up for publishing, not showing other people's reposts on your post.

dshanske:
The goal is to create a good default presentation things could override.

pfefferle:
But then we should at least stick with the default, so that there are no CSS differences (see my screenshot).

dshanske:
> But then we should at least stick with the default, so that there are no CSS differences (see my screenshot).

But the html5_comment version I included has no differences on the CSS by design. It adds microformats classes and the citation line that we had in Semantic Linkbacks.

dshanske:
Everything in this implementation, as well as the new comment implementation for 5.0 was designed to be overridden the way it always was. The new way of showing the reactions loads them, then loads the theme's comment display afterward. The comment walker overrides start_el in order to allow loading additional callbacks and additional options to wp_list_comments so it is backward compatible. It allows the theme to still use its own callback...

pfefferle:
@dshanske 

> Oops, I was in the wrong PR, ignore my "Approve", I want to wait, what @dshanske thinks about it!

I do not know how to un-approve 😞 it was by mistake!

includes/class-comment.php
pfefferle: ok, thanks for clarification!

includes/class-comment-walker.php
```diff
@@ -108,7 +108,7 @@ public function start_el( &$output, $data_object, $depth = 0, $args = array(), $
 
 		$defaults = array(
 			'avatar_only' => false, // Implements an argument that will just output an avatar
-			'overlay'     => true, // Implements an argument that optionally overlays an icon on top of the profile image, applies only to the avatar only output
+			'overlay'     => ( 1 == get_option( 'webmention_reaction_overlay', 1 ) ), // Implements an argument that optionally overlays an icon on top of the profile image, applies only to the avatar only output
```
pfefferle, line 111:
```suggestion
			'overlay'     => (bool) get_option( 'webmention_reaction_overlay', true ), // Implements an argument that optionally overlays an icon on top of the profile image, applies only to the avatar only output
```

includes/class-comment-walker.php
```diff
@@ -170,4 +169,115 @@ protected function avatar_only( $comment, $depth, $args ) {
...
+		// Only call this local version for comments that are webmention based.
+		if ( 'webmention' !== get_comment_meta( $comment->comment_ID, 'protocol', true ) ) {
+			parent::comment( $comment, $depth, $args );
+			return;
+		}
+
+		$tag = ( 'div' === $args['style'] ) ? 'div' : 'li';
+
+		$cite  = apply_filters( 'webmention_cite', '<small>&nbsp;@&nbsp;<cite><a href="%1s">%2s</a></cite></small>' );
+		$url = get_url_from_webmention( $comment );
```
pfefferle, line 190:
```suggestion
		$cite = apply_filters( 'webmention_cite', '<small>&nbsp;@&nbsp;<cite><a href="%1s">%2s</a></cite></small>' );
		$url  = get_url_from_webmention( $comment );
```

includes/class-comment-walker.php
```diff
@@ -170,4 +169,115 @@ protected function avatar_only( $comment, $depth, $args ) {
...
+		} else {
+			$moderation_note = __( 'Your comment is awaiting moderation. This is a preview; your comment will be visible after it has been approved.', 'default' );
+		}
+		?>
+		<<?php echo $tag; ?> id="comment-<?php comment_ID(); ?>" <?php comment_class( $this->has_children ? 'parent' : '', $comment ); ?>>
+			<article id="div-comment-<?php comment_ID(); ?>" class="comment-body h-cite <?php echo $type; ?>">
+				<footer class="comment-meta">
+					<div class="comment-author vcard h-card u-author">
+						<?php
+						if ( 0 != $args['avatar_size'] ) {
```
pfefferle, line 212:
```suggestion
						if ( 0 !== $args['avatar_size'] ) {
```

includes/class-comment-walker.php
```diff
@@ -170,4 +169,115 @@ protected function avatar_only( $comment, $depth, $args ) {
...
+							echo get_avatar( $comment, $args['avatar_size'] );
+						}
+						?>
+						<?php
+						$comment_author = get_comment_author_link( $comment );
+
+						if ( '0' == $comment->comment_approved && ! $show_pending_links ) {
```
pfefferle, line 219:
```suggestion
						if ( '0' === $comment->comment_approved && ! $show_pending_links ) {
```

includes/class-comment-walker.php
```diff
@@ -170,4 +169,115 @@ protected function avatar_only( $comment, $depth, $args ) {
...
+					<em class="comment-awaiting-moderation"><?php echo $moderation_note; ?></em>
+					<?php endif; ?>
+				</footer><!-- .comment-meta -->
+
+				<div class="comment-content e-content p-name">
+					<?php comment_text(); ?>
+				</div><!-- .comment-content -->
+
+				<?php
+				if ( '1' == $comment->comment_approved || $show_pending_links ) {
```
pfefferle, line 265:
```suggestion
				if ( '1' === $comment->comment_approved || $show_pending_links ) {
```

includes/class-comment-walker.php
```diff
@@ -170,4 +169,115 @@ protected function avatar_only( $comment, $depth, $args ) {
...
+						<?php
+						// Allow arbitrary additions to comment metadata.
+						do_action( 'webmention_comment_metadata', $comment );
+						printf(
+							'<a class="u-url" href="%s"><time class="dt-published" datetime="%s">%s</time></a>',
+							esc_url( get_comment_link( $comment, $args ) ),
+							get_comment_time( DATE_W3C ),
+							sprintf(
+								/* translators: 1: Comment date, 2: Comment time. */
+								__( '%1$s at %2$s' ),
```
pfefferle, line 245:
```suggestion
								__( '%1$s at %2$s', 'webmention' ),
```

dshanske: This is actually from the original, so it should be ,'default'

includes/class-comment-walker.php
dshanske: Oddly, all of the == vs === are straight out of the original html5_comment.

dshanske:
@pfefferle @chrisbergr Added a FAQ to explain how the existing system works. While the html5_comment addition happens in this plugin, everything else I noted was already in version 5.

I think it needs it. If we expand to add a locate_template option, it would simply be another alternative.

chrisbergr:
> If WordPress changes the default behaviour and/or the CSS changes, it might break Webmention representation.

The default template I provided uses the original code from [WordPress](https://developer.wordpress.org/reference/classes/walker_comment/html5_comment/). That's why CSS changes should not be a problem. If they change something on their html structure, it is the same problem as with any other thing that works with templates. It is not a problem that we then offer an update of the standard template. If you have ever worked with WooCommerce and have overwritten there email templates: They solve this quite nicely by noticing you that there have been changes and your template is no longer up to date.

> Most of the themes do not rely on the comment walker template at all and implement their own representation using the comment callback. See the autonomie theme for example: `wp_list_comments( array( 'callback' => 'autonomie_comment', 'format' => '' ) );`.

This is not an issue for this topic, it is more general. If the theme uses a custom walker, the walker provided by this plugin won't be used at all.

includes/class-admin.php
```diff
@@ -390,6 +390,16 @@ public static function register_settings() {
 				'default'      => 1,
 			)
 		);
+		register_setting(
```
pfefferle, line 393: I am not sure if we should add more settings like this, if we work on templating! We should keep the settings clean and maybe work on different templates, a user can choose from!

pfefferle: Maybe we add a const instead, for the quick-win.

dshanske: How about a filter?

dshanske: I added a filter, and if we remove the setting, then we just leave the filter.

pfefferle: I would remove the settings anyways, otherwise I think we get a lot messages if we remove them, why they are gone. Atm. there is only one person having the issue, so I would prefer using only the filter and/or a constant.

dshanske: Removed setting.

dshanske:
@pfefferle Wanted to see about remerging this.

pfefferle:
The problem with this PR is, that it will always overwrite the comment representation of the theme.

dshanske:
No, it's the opposite, the theme will always overwrite it.

pfefferle:
But in what case will the comment-walker version be used then?

dshanske:
When the theme doesn't set a custom rendering, or when the theme opts to use the built in.

pfefferle:
It is not directly because of this PR, but I see an error (on block themes) in the comments section:

> Deprecated: File Theme without comments.php is deprecated since version 3.0.0 with no alternative available. Please include a comments.php template in your theme. in /var/www/html/wp-includes/functions.php on line 5583

Issue #284, opened 2020-08-22
Show entire content text if short enough


dshanske:
Related: https://github.com/pfefferle/wordpress-semantic-linkbacks/issues/41

"If the mention post is a note and/or its content text is short enough - say, < 140 chars :P - it'd be nice to show the content text inline."

dshanske:
Would suggest we consider this for #361 but otherwise, it seems like it might conflict with the other presentation.

pfefferle:
I think this issue is about the facepiles setting from SL, where you where able to decide if a type shows up as comment or as facepiles.

dshanske:
I think this is specifically for type mention though.

pfefferle:
Ah, this is a ticket created by you! There is another ticket about the setting! Sorry, forget what I commented ☺️

dshanske:
@pfefferle Actually, it was bringing over one created by @snarfed

pfefferle:
I see!

pfefferle:
I think for this one we should change the comment type to "comment" if the requirements are met, to not make the output logic more complicated?

snarfed:
Also related: https://github.com/pfefferle/wordpress-webmention/issues/182

edent:
At the moment, I get around this by doing:

```php
$text_length = mb_strlen( wp_strip_all_tags( html_entity_decode( $comment_text, ENT_QUOTES ) ) );

if ( $text_length >= 1024 ) {
	$comment_text = "<a href=\"$comment_url\">Read more</a>.";
} else {
	$comment_text = apply_filters( 'comment_text', get_comment_text(), $comment);
}
```

But that could also truncate the text if it was too long.

chrisbergr:
Maybe there should be an option to set from which character length the text is truncated? Or at least a filter.

dshanske:
Resolved via #378 by having them set as comments.

PR #378, opened 2023-04-12
Short Mentions

dshanske:
Addresses #284 

This converts mentions to comments if the content is short enough.

### DIFF
--- a/includes/Entity/class-item.php
+++ b/includes/Entity/class-item.php
@@ -282,7 +282,18 @@ class Item {
 	 * @return string
 	 */
 	public function get_response_type() {
-		return $this->response_type ? $this->response_type : 'mention';
+		$response_type = $this->response_type ? $this->response_type : 'mention';
+		// Reclassify short mentions as comments
+		if ( 'mention' === $this->response_type ) {
+			$text = $this->get_content();
+			if ( is_string( $text ) ) {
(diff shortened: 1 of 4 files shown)


pfefferle (changes requested):
This is the same code in three different classes, that means there is a lot of potential to break something if we only add changes to some of them, or to miss it on new handlers.

I would prefer this to be added in the item. maybe on the `to_commentdata_array` or on the `get_response_type` function.

dshanske:
> This is the same code in three different classes, that means there is a lot of potential to break something if we only add changes to some of them, or to miss it on new handlers.
> 
> I would prefer this to be added in the item. maybe on the `to_commentdata_array` or on the `get_response_type` function.

There is no get_response_type function. It is a magic getter. I agree with you, it probably makes more sense to move it upstream.

includes/Entity/class-item.php
```diff
@@ -282,7 +282,18 @@ public function get_content() {
 	 * @return string
 	 */
 	public function get_response_type() {
-		return $this->response_type ? $this->response_type : 'mention';
+		$response_type = $this->response_type ? $this->response_type : 'mention';
+		// Reclassify short mentions as comments
+		if ( 'mention' === $this->response_type ) {
```
pfefferle, line 287:
```suggestion
		if ( 'mention' === $response_type ) {
```

otherwise it can be empty/null?!?

PR #381, opened 2023-04-13
Remove Gravatar Code

dshanske:
Being as it seems the thought is that we'd be better served by having avatar storage and caching handled by a separate plugin, this removes gravatar caching code, which is unnecessary as it only worked if the webmention had an author email, which is statistically unlikely.

### DIFF
--- a/includes/class-avatar.php
+++ b/includes/class-avatar.php
@@ -14,21 +14,11 @@ class Avatar {
 	 */
 	public static function init() {
 		add_filter( 'pre_get_avatar_data', array( static::class, 'avatar_stored_in_comment' ), 30, 2 );
-		add_filter( 'get_avatar_data', array( static::class, 'anonymous_avatar_data' ), 30, 2 );
-
-		// All the default gravatars come from Gravatar instead of being generated locally so add a local default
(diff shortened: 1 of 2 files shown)


includes/class-avatar.php
```diff
@@ -170,7 +63,7 @@ public static function avatar_stored_in_comment( $args, $comment ) {
 		}
 
 		// Simple method to prevent broken images
-		$args['extra_attr'] .= sprintf( ' onerror="this.onerror=null;this.src=\'%1$s\';this.srcset=\'%1$s\';"', self::get_default_avatar() );
+		$args['extra_attr'] .= sprintf( ' onerror="this.onerror=null;this.src=\'%1$s\';this.srcset=\'%1$s\';"', plugin_dir_url( dirname( __FILE__ ) ) . 'assets/img/mm.jpg' );
```
pfefferle, line 66: I would recommend to use one of these https://github.com/pfefferle/wordpress-webmention/blob/main/webmention.php#L34 instead of `plugin_dir_url( dirname( __FILE__ ) )`.

pfefferle: hmm, I think there is no one for URL. Maybe adding the URL one and use it here.

pfefferle (approved):
LGTM

PR #383, opened 2023-04-13
Allow refreshing webmentions from the Bulk Actions menu

dshanske:
This allows the refreshing of webmentions from the Bulk Actions menu, assuming the original source is still available.

### DIFF
--- a/includes/class-admin.php
+++ b/includes/class-admin.php
@@ -28,6 +28,9 @@ class Admin {
 		add_filter( 'manage_edit-comments_columns', array( static::class, 'comment_columns' ) );
 		add_filter( 'manage_comments_custom_column', array( static::class, 'manage_comments_custom_column' ), 10, 2 );
 
+		add_action( 'bulk_actions-edit-comments', array( static::class, 'bulk_comment_actions' ), 10, 3 );
+		add_filter( 'handle_bulk_actions-edit-comments', array( static::class, 'bulk_comment_action_handler' ), 10, 3 );
+
 		add_filter( 'comment_row_actions', array( static::class, 'comment_row_actions' ), 13, 2 );
 		add_filter( 'comment_unapproved_to_approved', array( static::class, 'transition_to_approvelist' ), 10 );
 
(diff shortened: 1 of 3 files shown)


includes/class-admin.php
```diff
@@ -70,6 +73,59 @@ public static function manage_comments_custom_column( $column, $comment_id ) {
...
+	public static function bulk_comment_actions( $bulk_actions ) {
+		$bulk_actions['refresh_webmention'] = __( 'Refresh Webmention', 'webmention' );
+		return $bulk_actions;
+	}
+
+	/**
+	 * Add bulk action handler to comments
+	 *
+	 */
+	public static function bulk_comment_action_handler( $redirect_to, $doaction, $comment_ids ) {
```
pfefferle, line 88: this is still a lot of code for refreshing a comment... maybe we should move this to the class-comment.php file, so that we could call it from several places...

pfefferle: it is also hard to check if it works or not, maybe we could also add a test for that?

dshanske: I did add a refresh timestamp but agree the function to refresh should be broken off

dshanske:
This adds a wp cli entry and a function. How does it look so far?

pfefferle:
Oh nice! Nice idea!

dshanske:
@pfefferle What else do you think this needs?

includes/functions.php
```diff
@@ -550,3 +550,46 @@ function webmention_comment_form() {
 		load_template( $template );
 	}
 }
+
+/**
+ * Refresh an existing comment
+ *
+ * @param int|WP_Comment Comment object or ID.
+ * @return WP_Error|bool Return true or error object.
```
pfefferle, line 558: isn't it easier to only return "WP_Error" or bool?

includes/class-cli.php
```diff
@@ -82,6 +82,67 @@ public function meta( $args, $assoc_args ) {
...
+	 *
+	 *     $ wp webmention refresh --post=1
+	 *
+	 * @param array|null $args       The arguments.
+	 * @param array|null $assoc_args The associative arguments.
+	 *
+	 * @return void
+	 */
+	public function refresh( $args, $assoc_args ) {
+		if ( $assoc_args ) {
```
pfefferle, line 110: I would rewrite this code to fail fast.

```PHP
if ( !$assoc_args ) {
    WP_CLI::error( __( 'Please provide a comment ID or a post-id/permalink', 'webmention' ) );
}

File: includes/functions.php (hunk 2 above)
Comment (dshanske) [reply includes/functions.php:558]: Isn't that what we're returning?

Comment (pfefferle) [reply includes/functions.php:558]: There are a lot of `false` returns.

Comment (pfefferle) [reply includes/functions.php:558]: I mean either only WP_Error or only false, but not mixing it.

File: includes/class-cli.php [hunk 4]
```diff
@@ -82,6 +82,67 @@ public function meta( $args, $assoc_args ) {
...
+			$comment_id = isset( $assoc_args['comment'] ) ? $assoc_args['comment'] : null;
+			if ( $comment_id ) {
+				$return = webmention_refresh( $comment_id );
+				if ( $return ) {
```
pfefferle, line 114: should check for `! is_wp_error( $return )` or `true === $return`

pfefferle:
LGTM!

pfefferle (approved):

PR #385, opened 2023-04-16
Set created at time only for new comments.

### LINKED ISSUE #360: Webmention Created At
The webmention created at is overwritten on update.

dshanske:
Resolves #360 as this will only set this property if it is a new comment.

### DIFF
--- a/includes/class-cli.php
+++ b/includes/class-cli.php
@@ -223,7 +223,7 @@ class Cli extends WP_CLI_Command {
 						'protocol'              => 'webmention',
 						'avatar'                => "https://i.pravatar.cc/80?u={$index}",
 						'url'                   => 'https://example.org/canonical',
-						'webmention_created_at' => current_time( 'mysql', 1 ),
+						'webmention_last_modified' => current_time( 'mysql', 1 ),
 						'webmention_source_url' => 'https://example.org/source',
 					),
 				)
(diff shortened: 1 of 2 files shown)


pfefferle:
And what about updated_at?

dshanske:
We don't have an updated at.

pfefferle:
oh, this is only for posts.

pfefferle:
Do we use `webmention_created_at` anywhere?

dshanske:
Comment date is set based on published date of the Webmention.. Creation could therefore be different

pfefferle:
Isn't it technically not possible to set the `comment_date` to the date that comes with the Webmention ping, or do you think this should be the local `created_at`?

pfefferle:
I am not sure I got this!?!

dshanske:
So, the  comment date is set using the published date. The created date is meant to reflect when the comment was created. The difference being that they are not always the same time.

dshanske:
Still confused about this. This is the actual time the webmention was created. The comment date field is the published date of the post the webmention came from.

pfefferle:
But the difference should be minutes... do we use the meta date for something?

dshanske:
Maybe we are better off with modified and stop storing created.

pfefferle:
I am just curious!

dshanske:
We originally put it in with an idea of tracking updates.. but we don't store updated dates.

pfefferle:
That would make more sense to me! That we use the comment_date for created_at and have a meta for updated_at!

dshanske:
@pfefferle Switched to last_modified.

pfefferle (approved):
LGTM

Issue #359, opened 2023-03-30
Are source URLs getting URL-decoded?


snarfed:
Hi @dshanske @pfefferle! I'm seeing an odd issue with source URLs with URL-encoded `#` characters, eg https://fed.brid.gy/render?id=https%3A%2F%2Findieweb.social%2Fusers%2Fsnarfed%23likes%2F709275 . That page has a `u-like-of` with a full `p-author h-card`, with name and photo, but when WordPress receives it as a webmention source, Semantic-Linkbacks doesn't find that author at all.

However, if I double-URL-encode the `#` character, ie https://fed.brid.gy/render?id=https%3A%2F%2Findieweb.social%2Fusers%2Fsnarfed%2523likes%2F709275 , a webmention with that source URL works fine.

I know URLs with `#`s are awkward, even when URL-encoded, but the first source URL is working ok with other wm receivers, eg https://www.jvt.me/week-notes/2023/09/ (scroll down and expand _Interactions with this post_), so I suspect this is a bug in this plugin or Semantic-Linkbacks?

Thanks in advance!

pfefferle:
@snarfed might be perhaps an issue with the Mf2 parser, because it supports fragment-parsing.

pfefferle:
@snarfed is the author outside of the fragment?

snarfed:
The source URL doesn't contain a fragment, it contains `%23`, which happens to be an _encoded_ `#` character. I think the plugin(s) are decoding that part of the URL, but shouldn't be, since the form-encoded POST body shouldn't be URL-decoded. (I think?)

Ideally the plugins would leave that `%23` in the URL alone when fetching it and parsing mf2.

pfefferle:
@snarfed a single hash is a "fragment" in the terminology of the MF2 parser: https://github.com/microformats/php-mf2/blob/main/Mf2/Parser.php#L2302

pfefferle:
This is a really good question!

pfefferle:
And the content type is: `application/x-www-form-urlencoded` so I would also assume that it has to be decoded!

snarfed:
From @sknebel [in chat](https://chat.indieweb.org/dev/2023-03-30#t1680189079875700):

> for keys and vaues, percent-encode everything "except the ASCII alphanumeric, U+002A (*), U+002D (-), U+002E (.), and U+005F (_). "
[...]

PR #386, opened 2023-04-16
Update dupe check

dshanske:
This updates the dupe check parameters to search for dupes by matching on webmention_source_url, url, _crossposting_link or source_url in one query instead of multiple queries. 

webmention_source url = the original source url
url = the canonical url in the event of something like a Bridgy post
_crossposting_link for Salmon
source_url for ActivityPub

### DIFF
--- a/includes/class-receiver.php
+++ b/includes/class-receiver.php
@@ -528,40 +528,40 @@ class Receiver {
 		$fragment = wp_parse_url( $commentdata['target'], PHP_URL_FRAGMENT );
 		// Meta Query for searching for the URL
 		$meta_query = array(
-					'relation' => 'OR',
-					// This would catch incoming webmentions with the same source URL
-					array(
-						'key'     => 'webmention_source_url',
-						'value'   => $commentdata['comment_meta']['webmention_source_url'],
-						'compare' => '=',
-					),
+			'relation' => 'OR',
+			// This would catch incoming webmentions with the same source URL
+			array(
+				'key'     => 'webmention_source_url',
+				'value'   => $commentdata['comment_meta']['webmention_source_url'],
+				'compare' => '=',
+			),
 
-					// This should catch incoming webmentions with the same canonical URL for Bridgy
-					array(
-						'key'     => 'url',
-						'value'   => $commentdata['comment_meta']['webmention_source_url'],
-						'compare' => '=',
-					),
-					// check comments sent via salmon are also dupes
-					// or anyone else who can't use comment_author_url as the original link,
-					// but can use a _crossposting_link meta value.
-					// @link https://github.com/pfefferle/wordpress-salmon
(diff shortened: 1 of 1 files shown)


dshanske:
Makes me wonder if we shouldn't migrate to source_url to match activitypub.

includes/class-receiver.php
```diff
@@ -501,10 +501,7 @@ public static function disable_wp_check_dupes( $dupe_id, $commentdata ) {
 			return $dupe_id;
 		}
 
-		if (
-			( isset( $commentdata['comment_type'] ) && 'webmention' === $commentdata['comment_type'] ) ||
-			( isset( $commentdata['comment_meta'] ) && ! empty( $commentdata['comment_meta']['semantic_linkbacks_type'] ) )
-		) {
+		if ( ! empty( $commentdata['comment_meta']['protocol'] ) && 'webmention' === $commentdata['comment_meta']['protocol'] ) {
```
pfefferle, line 504: can we alway assume that the array-keys exist?

dshanske: This is a check that is only supposed to be triggered on a Webmention, but no issue with an existence check

includes/class-receiver.php
```diff
@@ -524,28 +521,50 @@ public static function check_dupes( $commentdata ) {
...
-			$args = array(
-				'post_id'    => $commentdata['comment_post_ID'],
-				'meta_key'   => 'webmention_source_url',
-				'meta_value' => $commentdata['comment_author_url'],
+			),
+		);
+
+		if ( ! empty( $fragment ) ) {
+			// Ensure that if there is a
+			$args['meta_query'][0][] = array(
```
pfefferle, line 564: this looks a bit tricky, if the array order changes for example. can we make this a bit more robust?

includes/class-receiver.php
pfefferle: yes, sorry. wasn't aware, that `empty` also checks for `isset`.

dshanske:
I don't think we do. Don't we have to decode them?

snarfed:
for reference: https://github.com/pfefferle/wordpress-webmention/issues/359#issuecomment-1498347387

pfefferle:
@dshanske not decode, but encode! Because they are saved encoded didn't they? `esc_url`??

dshanske:
This is how I confuse myself

pfefferle:
😊

I think we should add some tests, to see what is correct!

pfefferle:
I would love to use this, to add some tests. Maybe @snarfed has some "special" URLs for us ;)

dshanske:
@pfefferle Do you want to merge this and do the urlencode testing as a separate PR?

pfefferle (approved):
LGTM

PR #387, opened 2023-04-17
fix comment separation

pfefferle:
See Mastodon thread: https://mastodon.social/@pfefferle/110204640637414553

thanks @jeherve for debugging!


jeherve (commented):
I can confirm that this fixes the issue on my own site. Thanks for the quick fix! 👍

dshanske:
The reason why twentytwenty specifically doesn't work is that it uses its own custom comment walker. Maybe instead of falling back, we should only fall back if we detect that the wp_list_comments function was called with a custom comment walker? On the other hand, how would we do that?

pfefferle:
I would not use different solutions for different themes. Because that also means that sometimes the facepiles are above, sometimes below and that results in a lot of support questions and makes debugging more complicated!

pfefferle:
I would merge this, as quick fix and evaluate other options in the course of the templating task.

PR #388, opened 2023-04-17
fix #369

### LINKED ISSUE #369: Undefined array key 1 in class-response.php on line 203
I'm seeing lots of error log entries with:

`PHP Warning:  Undefined array key 1 in /home/blog/wp-content/plugins/webmention/includes/class-response.php on line 203`

https://github.com/pfefferle/wordpress-webmention/blob/216b1b9727437a3c3d44d51476e4c551a14a9f6e/includes/class-response.php#L203

Appears to be part of `get_header_links()` - but I don't fully understand what the function does.

pfefferle:
I can not reproduce it, but it does not hurt to add checks before accessing the array

### DIFF
--- a/includes/class-response.php
+++ b/includes/class-response.php
@@ -197,13 +197,22 @@ class Response {
 				$item   = array();
 				$pieces = explode( ';', $link );
 				$uri    = array_shift( $pieces );
+
 				foreach ( $pieces as $p ) {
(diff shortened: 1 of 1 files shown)


dshanske (approved):
PHP8 is so specific, best to check everything

PR #392, opened 2023-04-23
Update tests and make `u-url` attrib optional for post-type discovery

pfefferle:


pfefferle:
please review @snarfed and @dshanske

dshanske (approved):
Not necessarily a fan of this, but I see the necessity.

pfefferle:
Can you explain a bit more?

dshanske:
> Can you explain a bit more?

I just get nervous about regex matching sometimes. I'm wondering if DOMDocument might be more consistent with what we did elsewhere. But it isnt' an issue.

pfefferle:
Merge this and think about a better way to get links in the content later?

dshanske:
I think so

PR #397, opened 2023-06-03
Add mf2 author migration

dshanske:
This bumps the schema to 1.0.1 in order to run a function to extract the author url from mf2_author and update author_url.

### DIFF
--- a/includes/class-db.php
+++ b/includes/class-db.php
@@ -8,7 +8,7 @@ class DB {
 	 *
 	 * @var int
 	 */
-	private static $target_version = '1.0.0';
+	private static $target_version = '1.0.1';
 
 	public static function get_target_version() {
 		return self::$target_version;
@@ -41,8 +41,9 @@ class DB {
 
 		$version_from_db = self::get_version();
 
-		if ( version_compare( $version_from_db, '1.0.0', '<' ) ) {
+		if ( version_compare( $version_from_db, '1.0.1', '<' ) ) {
 			self::migrate_to_1_0_0();
+			self::mf2_author_migration();
 		}
 
 		update_option( 'webmention_db_version', self::$target_version );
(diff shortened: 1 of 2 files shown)


includes/class-db.php
```diff
@@ -41,8 +41,9 @@ public static function update_database() {
 
 		$version_from_db = self::get_version();
 
-		if ( version_compare( $version_from_db, '1.0.0', '<' ) ) {
+		if ( version_compare( $version_from_db, '1.0.1', '<' ) ) {
```
pfefferle, line 44: Please add a different block for 1.0.1 and keep the 1.0.0, so that we can see what changed over time.

includes/class-db.php
```diff
@@ -114,4 +115,40 @@ public static function migrate_to_1_0_0() {
...
+		);
+		foreach ( $comments as $comment_id ) {
+			$author = get_comment_meta( $comment_id, 'mf2_author', true );
+			if ( is_array( $author ) ) {
+				if ( array_key_exists( 'url', $author ) ) {
+					$comment               = get_comment( $comment_id, ARRAY_A );
+					$comment['author_url'] = $author['url'];
+					wp_update_comment( $comment );
+				}
+				delete_comment_meta( $comment_id, 'mf2_author' );
```
pfefferle, line 150: Let's do not delete things for now ☺️

Maybe in a later version if we are sure we have anything migrated successfully!

dshanske: That's fine. Otherwise, does this look like a good idea or should I make some additional checks? Like comparing the URL to the source_url?

includes/class-db.php
```diff
@@ -44,6 +44,9 @@ public static function update_database() {
 		if ( version_compare( $version_from_db, '1.0.0', '<' ) ) {
 			self::migrate_to_1_0_0();
 		}
+		if ( version_compare( $version_from_db, '1.0.1', '<' ) ) {
+			self::mf2_author_migration();
```
pfefferle, line 48: Should we call it consistent with the 1_0_0?

pfefferle (approved):

PR #398, opened 2023-06-03
Include spam and trash statuses for dupe check

### LINKED ISSUE #341: If an existing webmention is in trash or spam, resending it shouldn't duplicate it
Right now, if a webmention is received, moved to Trash or Spam, and then gets resent (ie re-received), the plugin makes an entirely new copy of it as a new comment. It should update (if necessary) the existing comment in Trash or Spam instead. Background in https://github.com/snarfed/bridgy/issues/1338

cc @dshanske. Thanks in advance, hope you all are good!

dshanske:
Resolves #341

The code shows 'any' as an option, but the documentation doesn't. Filed a ticket. If I read WP_Comment_query correctly, all actually doesn't include spam and trash, but any does.

### DIFF
--- a/includes/class-receiver.php
+++ b/includes/class-receiver.php
@@ -567,6 +567,7 @@ class Receiver {
 		$args = array(
 			'post_id'    => $commentdata['comment_post_ID'],
 			'meta_query' => array( $meta_query ),
+			'status' => 'any'
 		);
 
 		if ( ! empty( $fragment ) ) {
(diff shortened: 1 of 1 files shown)


dshanske:
https://github.com/WordPress/wordpress-develop/blob/6.2/src/wp-includes/class-wp-comment-query.php#L555 I am reading this correctly, aren't I?

pfefferle (approved):
I read the same in it!

Issue #404, opened 2023-06-13
Comment form issue using block theme


janboddez:
After updating to v5.1.0, I get this error in my debug.log:
```
File Theme without comments.php is deprecated since version 3.0.0 with no alternative available. Please include a comments.php template in your theme.
```
I'm also suddenly seeing the comment form _twice_.

Disabling the plugin fixes both. I'm 99% sure it has to do with me using a _block theme_.

janboddez:
The Gutenberg dynamic comments block has this here code in it, to, I believe, work around the notice.
```
/*
	 * There's a deprecation warning generated by WP Core.
	 * Ideally this deprecation is removed from Core.
	 * In the meantime, this removes it from the output.
	 */
	add_filter( 'deprecated_file_trigger_error', '__return_false' );
	comments_template();
	remove_filter( 'deprecated_file_trigger_error', '__return_false' );
```
I've been looking for the code that may be calling `comments_template()`, can't readily find it. I'll try to dig a bit deeper.

janboddez:
So, everything works if I just disable "Separate Webmention Types from Comments." Which I can, because I'm doing this separately anyhow.

Update: Could be this is the issue: https://github.com/pfefferle/wordpress-webmention/blob/6e4ed42bea05966ba85e4b05399348ea382ea951/templates/webmention-comments.php#L37

Block themes don't have a `comments.php`.

A fix _might_ be (just guessing here) to instead call code above, same way Gutenberg does it.

janboddez:
Thinking it might habe been introduced here: https://github.com/pfefferle/wordpress-webmention/commit/5d41744fe7405a5d723dc167646b72050923e3c3

Webmention used to filter the `comment_template` (which I believe block themes _normally_ don't call, at all). Whereas now, the template seems to be called always (through `comment_form_before`).

janboddez:
OK, so what's happening is the `core/post-comments-form` block calls `comment_form()` which calls `comment_form_before` which loads the facepile template, which calls `locate_template( 'comments.php' )`.

But because block themes don't have it, it instead returns `wp-includes/theme-compat/comments.php`, which calls ... `comment_form()`.

I don't yet understand why "classic" themes don't suffer from this.

janboddez:
Looks like I'm not the only one experiencing this: https://cagrimmett.com/week-of/2023/06/11/week-of-june-5-2023/

janboddez:
The following (in `webmention-comments.php`) works for me:
[...]

PR #405, opened 2023-06-14
Remove unnecessary loading of `comments.php`

janboddez:
As mentioned in #404, I'm 95% sure `comments.php` should no longer be loaded at the end of the `comment_form_before` callback, precisely because the hook is called _from_ (a function inside) `comments.php`.

On "classic" sites, this somehow works (i.e., there is no noticeable problem), but on sites running a block theme, this leads to both a deprecation notice (expected behavior) and the comment form being shown twice.

(It might also run OK on "block theme sites" that utilize the so-called legacy comments block. Either way, I think it's safe to remove the call. I tested in combination with a number of block and classic themes from various vendors.)

### DIFF
--- a/templates/webmention-comments.php
+++ b/templates/webmention-comments.php
@@ -33,6 +33,3 @@ foreach ( $grouped_mentions as $mention_type => $mentions ) {
 }
 
 do_action( 'webmention_after_reaction_list' );
-
-load_template( locate_template( 'comments.php' ) );
-?>


dshanske:
The original 5.X version loaded our webmention display, then loaded the comment template in the theme. When that created a problem, @pfefferle switched it to the older way we did it in 4.X, which was to load the webmentions using the comment_form_before.

So, this is just the result of us missing something when we switched methods.

pfefferle (approved):
LGTM

Issue #399, opened 2023-06-06
<br /> not converted to \n from Bridgy


edent:
This comment - https://brid.gy/comment/mastodon/@edent@mastodon.social/110491395940578642/110495628806946433 contains the following:

```html
<a href="https://mastodon.social/@dracos" class="u-url mention">@<span>dracos</span></a></span><br />Lovely!</p>
```

But it appears in my comments without the `<br />` nor is it converted to a `\n`

![Screenshot of the comment.](https://github.com/pfefferle/wordpress-webmention/assets/837136/d5de0891-d893-482f-9615-954ac70ec969)

dshanske:
We do have a sanitize function, but it includes <br>. https://github.com/pfefferle/wordpress-webmention/blob/main/includes/functions.php#L382

Your link shows rel="no-follow ugc", so let's go through all the things WordPress does to filter comment content....doing this as an exercise to decide which ones we may wish to override.

Both of these are hooked to pre_comment_content, when things are added to the database.

- wp_rel_ugc - Adds rel="nofollow ugc" string to all HTML A elements in content.
- wp_filter_post_kses - Runs it through the post filter, rather than our custom filter.

Considering we filter, I don't think there's an issue with removing the default WordPress filtering. but it also allows br. 

@pfefferle Thoughts here?

snarfed:
On a related note, I'd encourage you all to think about this in terms of "how do we preserve the original reply's whitespace and formatting intact?" instead of "which tags do we sanitize or not?" Whitespace handling is (maybe obviously) both surprisingly difficult and surprisingly important, I've spent way more time on it in Bridgy etc than I ever expected to.

dshanske:
I think I was starting with should we stop letting WordPress sanitize it after we have? That's a double cleanup.

janboddez:
I'm _guessing_ that what happens is this: https://github.com/WordPress/wordpress-develop/blob/ba9a2f8b83211fbdee1621fd4f39dcb11908b817/src/wp-includes/kses.php#L2202. Or this: https://github.com/WordPress/wordpress-develop/blob/ba9a2f8b83211fbdee1621fd4f39dcb11908b817/src/wp-includes/comment.php#L3618. :-)

Either way, it looks like that, for webmention comments, there is no logged-in user; they're a back-end-only thing. Right? :-S

[...]

PR #408, opened 2023-07-15
Remove filter because we are already filtering

dshanske:
#399 This removes the filtering before adding the webmention, on the basis we do our own filtering in webmention_sanitize_html and restores it afterward.

### DIFF
--- a/includes/class-receiver.php
+++ b/includes/class-receiver.php
@@ -351,8 +351,18 @@ class Receiver {
 				$commentdata['comment_meta'] = array();
 			}
 
-			// save comment
+			// save comment but remove content filtering because we filter our own content
+			remove_filter( 'pre_comment_content', 'wp_filter_post_kses' );
+			remove_filter( 'pre_comment_content', 'wp_filter_kses' );
+
 			$commentdata['comment_ID'] = wp_new_comment( $commentdata, true );
+
+			// restore filter after add
+			if ( current_user_can( 'unfiltered_html' ) ) {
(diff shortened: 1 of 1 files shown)


pfefferle (commented):
Shouldn't we instead remove our kses? Isn't that easier?

includes/class-receiver.php
```diff
@@ -351,9 +351,19 @@ public static function post( $request ) {
...
 
-			// save comment
+			// save comment but remove content filtering because we filter our own content
+			remove_filter( 'pre_comment_content', 'wp_filter_post_kses' );
+			remove_filter( 'pre_comment_content', 'wp_filter_kses' );
+
 			$commentdata['comment_ID'] = wp_new_comment( $commentdata, true );
 
+			// restore filter after add
+			if ( current_user_can( 'unfiltered_html' ) ) {
```
pfefferle, line 361: This is an API request, so there is no current_user!?!

dshanske:
> Shouldn't we instead remove our kses? Isn't that easier?

#399 suggests that the built-in WordPress filtering is taking out elements we would keep in.

includes/class-receiver.php
pfefferle: what if there is no current_user? will it fail gracefully?

dshanske: Yes.

pfefferle (approved):
LGTM, thanks @dshanske

PR #409, opened 2023-07-17
Switch from set to add to avoid meta overriding mf2

dshanske:
This is based on the webmention sent from https://minutestomidnight.co.uk/notes/2023-07-13-10-52-06/ to 
https://tracydurnell.com/2023/07/12/barriers-to-a-more-social-indieweb/ where the content is actually wrong and coming from meta when a better microformats options is there.


pfefferle (approved):
Makes sense!

PR #411, opened 2023-07-30
Use summary when appropriate

dshanske:
If the response type is expressly set to comment, then this will always return the full content. Otherwise, it will return the content if it is short enough, if not the summary, if not the name, if not nothing.

I debated on the nothing vs a generated excerpt. But if we were to do a generated excerpt, I'd want to discuss how we'd excerpt as a topic. For example, number of characters = MAX_INLINE_MENTION_LENGTH? Try to mirror Pingbacks and get the context around the link to the target? Etc.

### DIFF
--- a/includes/Entity/class-item.php
+++ b/includes/Entity/class-item.php
@@ -285,9 +285,9 @@ class Item {
 		$text_len = $this->str_length( $this->content );
 
 		if ( ( 0 === $text_len ) ) {
-			if ( ! empty( $this->summary ) ) {
+			if ( $this->summary ) {
 				return $this->summary;
-			} elseif ( ! empty( $this->name ) ) {
+			} elseif ( $this->name ) {
 				return $this->name;
 			}
 		}
@@ -298,9 +298,9 @@ class Item {
 		}
 
 		// If there is a summary that isn't empty, return that, if not the name.
-		if ( ! empty( $this->summary ) ) {
+		if ( $this->summary ) {
 			return $this->summary;
-		} elseif ( ! empty( $this->name ) ) {
+		} elseif ( $this->name ) {
 			return $this->name;
 		}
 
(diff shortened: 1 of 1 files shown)


includes/Entity/class-item.php
```diff
@@ -282,15 +282,29 @@ public function get_author( $property = null ) {
...
+
+		// If there is a summary that isn't empty, return that, if not the name.
+		if ( ! empty( $this->summary ) ) {
 			return $this->summary;
-		} elseif ( $this->name ) {
-			return $this->name;
-		} else {
-			return '';
+		} elseif ( ! empty( $this->name ) ) {
+			return $this - name;
```
pfefferle, line 304: @dshanske this should be `$this->name`?!!

pfefferle (changes requested):

includes/Entity/class-item.php
dshanske: Yes, phpcs...

includes/Entity/class-item.php
```diff
@@ -282,15 +282,29 @@ public function get_author( $property = null ) {
 	 * @return string
 	 */
 	public function get_content() {
-		if ( $this->content ) {
+		$text_len = $this->str_length( $this->content );
+
+		if ( ( 0 === $text_len ) ) {
+			if ( ! empty( $this->summary ) ) {
```
pfefferle, line 288: It think `if ( $this->summary ) {` should do the exact same and maybe a bit easier to read?!?

dshanske: Okay. I'll put it back to just $this->summary, but what if it isn't a string? Probably should check that too.

pfefferle: is that really possible?

dshanske:
Do you think this will address the raised behavior issue?

pfefferle (approved):

PR #412, opened 2023-07-30
some improvements and removed unused code

pfefferle:


dshanske:
We're not going to sanitize it using our function?

pfefferle:
The code I removed is never be used! Besides of that, we have to check sanitisation! Best practice is sanitize as late as possible, so we should add it at to_array and to_commentarray (some of the attribs are not Esche/sanitized at all!)

pfefferle:
Or we have to update the setters adders to also sanitize

dshanske:
Good point about moving it.

dshanske (approved):

PR #413, opened 2023-08-02
Version Bump to 5.1.2

dshanske:


pfefferle (commented):
I would prefer a minor, there are a lot of changes that are bigger than a simple patch. What do you think?

pfefferle (approved):

PR #416, opened 2023-09-18
Performance: only enqueue CSS on pages with a comment form

jeherve:
Instead of enqueuing the Webmention CSS on all pages of the site, we could try only enqueuing it on singular pages (posts / pages / CPTs) where comments are open.


pfefferle (approved):
LGTM!

PR #418, opened 2023-09-25
fix threaded comments

### LINKED ISSUE #414: Webmention Comment uses wrong link in reply-to target
I notice that a Webmention comment _to a comment_ never reached it's target. When trying to submit this comment manually I got informed that the target URL is not found. On closer inspection I confirmed that `webmention-comment.php` assigns the comment_author_url (e.g. `https://david.shanske.com/`) when it should use the _source_ URL of the comment I'm replying to: `https://david.shanske.com/2023/08/07/6935/`:

image

Made a little change (not having an idea about side effects) to fix this:

```php
// check parent comment
if ( $comment->comment_parent ) {
	// get parent comment...
	$parent = get_comment( $comment->comment_parent );
	// ...and gernerate target url
	if ( $parent->comment_author_url ) {
		$target = $parent->comment_author_url;
	}

	$source_url = get_comment_meta( $parent->comment_ID, 'webmention_source_url', true );
	if($source_url && !empty($source_url)) {
		$target = $source_url;
	}
}
```

Now I can submit `https://beko.famkos.net/2023/08/07/15628749/?replytocom=8719` to `https://david.shanske.com/2023/08/07/6935/` and don't get the error "No target found". Now @dshanske site throws an error 500 but that's probably another issue :zany_face:

pfefferle:
fix #414


pfefferle:
does that fix the issue for you? @bekopharm

dshanske (approved):
Looks good to me.

PR #421, opened 2023-09-26
Client-side URl validation

### LINKED ISSUE #420: Add "required" attribute for HTML form - and some validation?
If a user doesn't enter a URl and then presses the submit button, they get a strange error message:

> Invalid parameter(s): source

By adding the [`required`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required) attribute to the `<input>`  the user won't be able to submit a blank form and will be less likely to see an error.

Additionally, `<input type="url">` allows for *any* URl - `mailto:`, `ftp://`, `irc:` etc.

This can be achieved by using `pattern="^https?:\/\/(.*)"`

So the new input will look like:

```html
<input id="webmention-source"
             type="url"
             autocomplete="url"
             required
             pattern="^https?:\/\/(.*)"
             name="source"
             placeholder="<?php esc_attr_e( 'URL/Permalink of your article', 'webmention' ); ?>" />
```

edent:
Fixes #420

### DIFF
--- a/templates/webmention-comment-form.php
+++ b/templates/webmention-comment-form.php
@@ -9,7 +9,7 @@ do_action( 'webmention_comment_form_template_before' );
 		<label for="webmention-source"><?php echo get_webmention_form_text( get_the_ID() ); ?></label>
 	</p>
 	<p>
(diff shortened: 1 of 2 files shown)


pfefferle:
Isn't `type="url"` already testing for a valid URL?

pfefferle:
Ah, read your issue afterwards ☺️

pfefferle (approved):
LGTM

PR #426, opened 2023-10-21
Replace pings_open with new wrapper function, webmentions_open 

dshanske:
This moves the decision as to whether webmentions are open to a new webmentions_open function.

For this first phase, it still checks pings_open, so there is effectively no change. But this allows for changing this behavior in a future round.


dshanske:
The goal ultimately is to move to a system by which webmentions are open for all of a post type that is declared with webmention support unless expressly disabled by using the new per post flag.

pfefferle (approved):

PR #428, opened 2023-10-22
Add Optional Avatar Store

dshanske:
This brings back the Local Avatar Store code we wrote two years ago, in a minimalist manner, and it adds it to the repo disabled by default.

The way it is set up, as you may recall(as you co-wrote it), since it updates the Avatar property, if you disable it, it will retain for the existing updated ones, but it won't revert to remote loading until the next time the webmention is updated.

### DIFF
--- a/includes/class-avatar-store.php
+++ b/includes/class-avatar-store.php
@@ -1,0 +1,197 @@
+<?php
+
+namespace Webmention;
+
+use WP_Comment;
+
+/**
+ * Avatar Store Class
+ *
+ */
+class Avatar_Store {
+	/**
+	 * Initialize the plugin, registering WordPress hooks.
+	 */
+	public static function init() {
+		// Store Avatars Locally
+		add_action( 'comment_post', array( static::class, 'store_avatar' ), 20 );
+		add_action( 'edit_comment', array( static::class, 'store_avatar' ), 20 );
+		// add_action( 'get_comment', array( static::class, 'store_avatar' ), 20 );
+	}
+
+	/**
+	 * Return upload directory.
+	 *
+	 * @param string $filepath File Path. Optional
+	 * @param boolean $url Return a URL if true, otherwise the directory.
+	 * @return string URL of upload directory.
+	 */
+	public static function upload_directory( $filepath = '', $url = false ) {
+		$upload_dir  = wp_get_upload_dir();
+		$upload_dir  = $url ? $upload_dir['baseurl'] : $upload_dir['basedir'];
+		$upload_dir .= '/webmention/avatars/';
+		$upload_dir  = apply_filters( 'webmention_avatar_directory', $upload_dir, $url );
+		return $upload_dir . $filepath;
+	}
+
+	/**
+	 * Sideload Avatar
+	 *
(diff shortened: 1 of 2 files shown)


includes/class-avatar-store.php
```diff
@@ -0,0 +1,197 @@
...
+ */
+class Avatar_Store {
+	/**
+	 * Initialize the plugin, registering WordPress hooks.
+	 */
+	public static function init() {
+		// Store Avatars Locally
+		add_action( 'comment_post', array( static::class, 'store_avatar' ), 20 );
+		add_action( 'edit_comment', array( static::class, 'store_avatar' ), 20 );
+		// add_action( 'get_comment', array( static::class, 'store_avatar' ), 20 );
```
pfefferle, line 19: Why is this commented out?

dshanske: Because it led to excessive slowdowns in testing, so I decided to start with just when created or updated. Should have removed the line entirely.

dshanske:
@pfefferle Do you think I should bundle this into the next release...with all the fixes...or wait?

includes/class-avatar-store.php
```diff
@@ -0,0 +1,196 @@
...
+					'd' => $default, // Replace with our site default.
+					'r' => $rating,
+				),
+				$url
+			);
+			$file = download_url( $url, 300 );
+			if ( is_wp_error( $file ) ) {
+				return false;
+			}
+			@move_uploaded_file( $file, $filepath );
```
pfefferle, line 87: please don't mute errors. what could possibly go wrong here?

dshanske: You wrote that line... but I'll see about updating it.

pfefferle: is this my draft without changes? 😳

dshanske: This part is untouched from your original draft.

I mostly simplified the logic and the hooks.. but the storage and sideload code was as in your draft.

dshanske: Also, to be fair, the silencing of errors is done in the same way in WordPress core for the sideload functions.

pfefferle: then let's keep it 🫤

dshanske:
@pfefferle So what do you think? It is still disabled by default for now.

webmention.php
```diff
@@ -21,6 +21,11 @@
 defined( 'WEBMENTION_COMMENT_TYPE' ) || define( 'WEBMENTION_COMMENT_TYPE', 'webmention' );
 defined( 'WEBMENTION_GRAVATAR_CACHE_TIME' ) || define( 'WEBMENTION_GRAVATAR_CACHE_TIME', WEEK_IN_SECONDS );
 
+
+defined( 'WEBMENTION_LOCAL_AVATAR_STORE' ) || define( 'WEBMENTION_LOCAL_AVATAR_STORE', false );
```
pfefferle, line 25: should we use `WEBMENTION_AVATAR_LOCAL_STORE` to be consistent with the other Avatar Consts?

pfefferle (approved):

PR #429, opened 2023-10-28
Introduce HTML link search and add it to wp as a fallback

dshanske:
This adds functions to retrieve HTML links in the response class and then uses it in the wp search if it can't find an API link in the header.

### DIFF
--- a/includes/class-response.php
+++ b/includes/class-response.php
@@ -4,6 +4,7 @@ namespace Webmention;
 
 use WP_Error;
 use DOMDocument;
+use DOMXPath;
 
 /**
  * Encapsulates all Webmention HTTP requests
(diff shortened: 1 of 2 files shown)


includes/class-response.php
```diff
@@ -231,6 +232,27 @@ public function get_header_links() {
...
+	 */
+	public function get_html_links() {
+		$dom = $this->get_dom_document();
+		$xpath = new DOMXPath( $dom );
+		$items = array();
+
+		// check <link> and <a> elements
+		foreach ( $xpath->query( '(//link|//a)[@rel and @href]' ) as $link ) {
+			$item = array();
+			$item['rel']   = $link->getAttribute( 'rel' );
```
pfefferle, line 248: Intentions

dshanske: What do you mean by intentions? I'm mirroring the header link extraction function? Technically, it extracts all rels and then matches them against the passed parameters?

pfefferle: Oops, typo! Indents! 🫣

dshanske: Oh...I thought you were questioning why I did it. I really always forget to recheck after I run PHPCS Auto-fix. It sometimes messes up my indents as much as it fixes them.

dshanske: But those indents look right to me, so maybe I'm still indent-challenged.

pfefferle: The equal sign? Maybe a glitch on my phone...

dshanske: Better?

pfefferle (approved):

Issue #430, opened 2023-10-30
Handle time zones in published/updated mf2 values


snarfed:
It looks like time zones in `published` microformats2 values are currently ignored...? Looking at @sknebel's ("Anonymous"'s) and Scott Jenson's webmentions on https://snarfed.org/2023-10-30_51256#comments, their source mf2 both have `dt-published` values in UTC, ie they end with `+00:00`, but the plugin ignored that and uses their times directly in my WordPress's local time zone of `Los Angeles` (PT).

Specifically, here are the two `dt-published` values and source URLs:
* https://www.svenknebel.de/posts/2023/10/4/ , `2023-10-30 14:13:12.781417+00:00`
* https://ap.brid.gy/convert/web/https://social.coop/users/scottjenson/statuses/111324530456943368 , `2023-10-30T14:52:07+00:00`

I don't know whether this regressed in 5.x or time zones were never supported at all. #415 is loosely related. Thanks in advance!

dshanske:
The code pulls in the date and then formats it to store on the WordPress database.. which omits the timezone offset... but forgot to convert it to the site timezone.

dshanske:
Resolved via #432

PR #432, opened 2023-10-31
Repair Errors in the MF2 File

dshanske:
This adds a function to retrieve the author URL to get the name, photo, etc when it isn't embedded in the page and it fixes a PHP bug when something isn't a numeric array.


dshanske:
Added one more tweak, removed the get_published_gmt and add timezone as an argument to get_published. Also fixed the issue where the timestamp isn't set to the website timezone(#430)

pfefferle (approved):
Looks good, but I have not tested it

PR #437, opened 2023-11-06
simplify html/head link discovery

pfefferle:
Simplify the discovery mechanism by introducing a `get_links_by` function in the Response Object, that checks HTML and HTTP-Header Links.


dshanske (approved):
Looks good, but didn't test

PR #439, opened 2023-11-17
fix comment_link issue

pfefferle:
See https://wordpress.org/support/topic/breaks-ordinary-comment-permalink-links/

### DIFF
--- a/includes/class-comment.php
+++ b/includes/class-comment.php
@@ -31,7 +31,13 @@ class Comment {
 	 * @return string $url The url of the source.
 	 */
 	public static function remote_comment_link( $comment_link, $comment ) {
-		return get_url_from_webmention( $comment );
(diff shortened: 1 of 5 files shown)


includes/class-comment.php
```diff
@@ -31,7 +31,13 @@ public static function init() {
 	 * @return string $url The url of the source.
 	 */
 	public static function remote_comment_link( $comment_link, $comment ) {
-		return get_url_from_webmention( $comment );
+		$link = get_url_from_webmention( $comment );
+
+		if ( $link ) {
+			$comment_link = $link;
```
dshanske, line 37: Shouldn't it be just return $link at this point?

dshanske (approved):
Made a single suggestion.

PR #440, opened 2023-11-18
use source-link only on frontend and for Webmentions

pfefferle:
See: https://wordpress.org/support/topic/breaks-ordinary-comment-permalink-links/


dshanske (approved):
Looks good to me. Is there any scenario where we'd want the comment link in the admin though?

pfefferle:
Because of maybe put the check directly to the `get_url_from_webmention` function?

PR #443, opened 2023-12-22
Do not bypass spam detection

pfefferle:
In the current code, we bypass spam detection completely, because every Webmentions is always only `1` or `0`.

That makes spam detection useless and with spammers keeping up with new technologies like Webmentions, we should not bother our users to check every Mention.

What do you think @dshanske 

/cc @kbrown9 and @cybeardjm

### DIFF
--- a/includes/class-receiver.php
+++ b/includes/class-receiver.php
@@ -684,7 +684,7 @@ class Receiver {
 			return 1;
 		}
 
-		return self::is_source_allowed( $commentdata['source'] ) ? 1 : 0;
+		return self::is_source_allowed( $commentdata['source'] ) ? 1 : $approved;
 	}
 
 	/**
--- a/languages/webmention.pot
+++ b/languages/webmention.pot
@@ -2,10 +2,10 @@
 # This file is distributed under the MIT.
 msgid ""
 msgstr ""
-"Project-Id-Version: Webmention 5.1.9\n"
+"Project-Id-Version: Webmention 5.1.10\n"
 "Report-Msgid-Bugs-To: "
 "https://wordpress.org/support/plugin/wordpress-webmention\n"
-"POT-Creation-Date: 2023-11-18 15:05:08+00:00\n"
+"POT-Creation-Date: 2023-12-22 13:20:41+00:00\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=utf-8\n"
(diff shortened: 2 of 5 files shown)


dshanske (changes requested):
I'm sort of mixed on this. On one hand, letting WordPress make the decision on this is respecting the system. We have no need to do all moderation inside the plugin. Conversely though, no spam system so far is really equipped to make decisions on webmentions because they don't know what they are. But, the side of not superseding other plugins on the matter seems to win out.

dshanske (approved):

pfefferle:
> Conversely though, no spam system so far is really equipped to make decisions on webmentions because they don't know what they are.

That's not completely true, this request came initially from the Akismet team and we are working on supporting Pings in the next version of Antispam Bee!

When we always set all Webmentions either of `1` or `0` Akismet or Antispam Bee have no chance to train their algorithms from false positives for example.

dshanske:
> > Conversely though, no spam system so far is really equipped to make decisions on webmentions because they don't know what they are.
> 
> That's not completely true, this request came initially from the Akismet team and we are working on supporting Pings in the next version of Antispam Bee!
> 
> When we always set all Webmentions either of `1` or `0` Akismet or Antispam Bee have no chance to train their algorithms from false positives for example.

That's why I approved.

pfefferle:
Ah, I only saw the red bubble 🫣

But wanted to give you the context anyway!

dshanske:
> Ah, I only saw the red bubble 🫣
> 
> But wanted to give you the context anyway!

Accidentally clicked the wrong response and fixed it.

cybeardjm:
Hi @pfefferle - Just updated to the latest version.
Now all pings (reposts, likes...) coming from Bridgy go directly to spam!

pfefferle:
@cybeardjm have you added "brid.gy" to the allow-list?

@cybeardjm I will talk to the Akismet Team, to see what we can do.

kbrown9:
Hi @pfefferle! @cybeardjm did reach out to us, and I'm investigating the issue.

pfefferle:
Oh even better, thanks @kbrown9

cybeardjm:
@pfefferle @kbrown9 Nope, didn't modify my allow-list since our last exchange (well, I just removed my own domain, to avoid having some "comments" auto-validated).

Issue #444, opened 2023-12-22
Interpreting Response Markup as a Response


dshanske:
@snarfed reported https://chat.indieweb.org/dev/2023-12-20/1703111490653600

Now, this code interprets that as a like because of the class mapper

https://github.com/pfefferle/wordpress-webmention/blob/main/includes/Handler/class-mf2.php#L595

repost, like, favorite, bookmark, read, listen, watch, reply are used to mark up incoming for display, example: https://indieweb.org/like#receive_and_markup_incoming_likes

So, we shouldn't use them in the mapping to represent our like type for example, because the source didn't like the target, the target liked the source.

The second problem here is how exactly we do interpret when someone sends us a webmention that  is in the responses part of a post? Because it is meant to be an indication that the thing @snarfed responded to has updated. How do we interpret that?

PR #445, opened 2023-12-22
Support Source Update Notifications

dshanske:
This addresses #444 and creates the 'update' comment type, indicating a source update notification. For now, it doesn't store it in the database, but it adds a hook for someone to add notifications or otherwise on it.

It also, as a precaution, adds a check for unregistered webmention comment types being added. This is to ensure if someone adds something custom to the map, they register it.

### DIFF
--- a/includes/Handler/class-mf2.php
+++ b/includes/Handler/class-mf2.php
@@ -595,17 +595,17 @@ class MF2 extends Base {
 	protected function get_class_mapper() {
 		$class_mapper = array();
 
-		/* 
+		/*
 		 * These classes represent the source sending a webmention based on its display of a mention sent by the target.
 		 * This sets the type to update, indicating that the purpose of the webmention is to notify of an update to the original webmention.
 		 */
-		$class_mapper['comment'] = 'update';
-		$class_mapper['like']    = 'update';
-		$class_mapper['favorite']    = 'update';
-		$class_mapper['repost']    = 'update';
-		$class_mapper['bookmark']    = 'update';
-		$class_mapper['read']    = 'read';
-		$class_mapper['listen']    = 'listen';
+		$class_mapper['comment']  = 'update';
+		$class_mapper['like']     = 'update';
+		$class_mapper['favorite'] = 'update';
+		$class_mapper['repost']   = 'update';
+		$class_mapper['bookmark'] = 'update';
+		$class_mapper['read']     = 'read';
+		$class_mapper['listen']   = 'listen';
 		$class_mapper['watch']    = 'watch';
 
 		/*
--- a/includes/Handler/class-mf2.php
+++ b/includes/Handler/class-mf2.php
@@ -595,6 +595,19 @@ class MF2 extends Base {
 	protected function get_class_mapper() {
 		$class_mapper = array();
 
+		/* 
+		 * These classes represent the source sending a webmention based on its display of a mention sent by the target.
+		 * This sets the type to update, indicating that the purpose of the webmention is to notify of an update to the original webmention.
+		 */
+		$class_mapper['comment'] = 'update';
+		$class_mapper['like']    = 'update';
+		$class_mapper['favorite']    = 'update';
+		$class_mapper['repost']    = 'update';
+		$class_mapper['bookmark']    = 'update';
+		$class_mapper['read']    = 'read';
+		$class_mapper['listen']    = 'listen';
+		$class_mapper['watch']    = 'watch';
+
 		/*
 		 * rsvp
 		 * @link https://indieweb.org/rsvp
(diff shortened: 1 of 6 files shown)


includes/Handler/class-mf2.php
```diff
@@ -595,6 +595,19 @@ protected function urls_match( $url1, $url2 ) {
 	protected function get_class_mapper() {
 		$class_mapper = array();
 
+		/*
+		 * These classes represent the source sending a webmention based on its display of a mention sent by the target.
+		 * This sets the type to update, indicating that the purpose of the webmention is to notify of an update to the original webmention.
+		 */
+		$class_mapper['comment']  = 'update';
```
pfefferle, line 602: Can you explain that a bit more? I am not sure I can follow your thoughts!

includes/Handler/class-mf2.php
```diff
@@ -595,6 +595,19 @@ protected function urls_match( $url1, $url2 ) {
...
+		$class_mapper['like']     = 'update';
+		$class_mapper['favorite'] = 'update';
+		$class_mapper['repost']   = 'update';
+		$class_mapper['bookmark'] = 'update';
+		$class_mapper['read']     = 'read';
```
pfefferle, line 607: Shouldn't this and the following two be also "update"?

includes/class-receiver.php
```diff
@@ -341,6 +341,44 @@ public static function post( $request ) {
...
+		/*
+		 * Rejects Adding Non-Registered Webmention Comment Types. Ensures any plugins that add extra handling register their comment types.
+		 */
+		if ( ! is_registered_webmention_comment_type( $commentdata['comment_type'] ) ) {
+			/**
+			 * Fires if an Unregistered Comment Type is About to Be Added
+			 *
+			 * @param array $commentdata
+			 */
+			do_action( 'webmention_unknown_type', $commentdata );
```
pfefferle, line 378: Shouldn't we simply fall back to "mention" in that case instead of ignoring the ping?

includes/Handler/class-mf2.php
pfefferle: We should also maybe prefix this with something, because there might be an `Update` type in the future?!?

includes/Handler/class-mf2.php
dshanske: In each of these cases, it found the source link on a property indicating a marked up response. So, it isn't mentioning the target...at least that isn't the intention.

includes/Handler/class-mf2.php
dshanske: Yes, the read is just my mistake not updating it when I moved it out of place. No issue with a prefix. How about source_update?

includes/class-receiver.php
dshanske: That's probably acceptable.

pfefferle: I thought a bit about that, there might never be the case for an unsupported type, because the handlers use mention as default so maybe we should keep your proposal because it seems to be a bug or misuse of a filter then!

includes/Handler/class-mf2.php
pfefferle: 🤔

So this case should only happen if someone has a post with reactions and updates that post?!?

dshanske: Correct.

includes/Handler/class-mf2.php
pfefferle: But then I would vote for a complete remove of them, because otherwise these posts will never be updated!?!

dshanske: They are removed...in the receiver, after triggering an action. They never get added to the database. The mapper needs some way to tell the receiver to remove them...so using the comment_type.

pfefferle: Sorry I was totally on the wrong path! Time for holidays!

So it is not (only) a parser problem, because it should have never sent the ping (because the url is a reaction and should not be pinged) and only because there where a ping, it was falsely parsed!

Makes also sense!

includes/Handler/class-mf2.php
pfefferle: What about something like `error` or `failure`, then we would not need to prefix it. `Update` sound like that it is an expected ping!?!

dshanske: Let's say that my Post A is a reply to Post B. Post B marks it up as a response. then Post B's author updates the post and sends a webmention to Post A. If Post A has a reply-context on it, then the webmention sent by Post B should update that reply context. We don't have that feature, and it wouldn't be webmentions doing it necessarily, but by not having the hook that acknowledges it...

It isn't an error so much as an unhandled feature right now.

pfefferle: You mean something like a reverse update? Kind of "you referenced my blog and I updated it, maybe you are interested in that change" like?

pfefferle: It's not a bug it's a (not yet defined/implemented) feature? ☺️

dshanske: Exactly. So...the solution is to have the action that someone could add the feature to, but otherwise for now discard the mention.

pfefferle: Makes also sense! Sorry for standing on the hose for so long!

But I think this would be a target_update then?!?

pfefferle (approved):
Thanks David and get well soon!

PR #450, opened 2023-12-27
Add in default comment types

dshanske:
When I had it check for comment types, I forgot the built-in ones.


pfefferle:
Shouldn't we register them as types for Webmention instead?

dshanske:
Only comments I suppose.

pfefferle (approved):

PR #451, opened 2023-12-28
Improve fallback handling

### LINKED ISSUE #393: HackerNews links don't work
Tried to submit https://news.ycombinator.com/item?id=35927509 to https://shkspr.mobi/blog/2023/05/the-new-zip-tld-is-going-to-cause-some-problems/

I got the error

> Not enough data available

pfefferle:
if properties are not set and to avoid "anonymous".

fix #393

### DIFF
--- a/includes/Entity/class-item.php
+++ b/includes/Entity/class-item.php
@@ -24,7 +24,7 @@ class Item {
 	protected $updated;
 
 	/**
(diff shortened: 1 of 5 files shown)


dshanske (approved):
Looks good to me. I have some future ideas, but this works for now.

PR #457, opened 2024-01-06
Fix #396

### LINKED ISSUE #396: Avatar overlaps text in Dashboard widget
There's a `<div class="dashboard-comment-wrap has-row-actions">` in there, which would need a `has-avatar` class ...

Not sure if I'm the only one seeing this?

The second mention looks okay because as far as WordPress is concerned, it's just a comment (its comment type is empty).

janboddez:
Adds a single CSS file on the admin dashboard only, which I believe makes the "custom" comment types look more like WordPress' default comments (in the dashboard's "Activity" panel).

Tested (only) on my own live site (for different comment types, with and without avatar). Seems to work :-P

### DIFF
--- a/assets/css/admin.css
+++ b/assets/css/admin.css
@@ -1,0 +1,27 @@
+.repost .dashboard-comment-wrap,
+.like .dashboard-comment-wrap,
+.favorite .dashboard-comment-wrap,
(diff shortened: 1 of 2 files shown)


pfefferle (approved):
LGTM; Thanks :)

PR #462, opened 2024-03-12
Comment style

dshanske:
Addresses https://wordpress.org/support/topic/plugin-strips-markup-from-non-webmention-comments


pfefferle:
Isn't there a setting we could check? (Like theme support)

dshanske:
> Isn't there a setting we could check? (Like theme support)

But why, if we are turning control over to the built-in functionality? And if the theme has a custom comment callback, it overrides everything?

pfefferle (approved):

Issue #438, opened 2023-11-07
Bridgy Webmentions Set Content to Bridgy Response


dshanske:
If the Bridgy webmention has no content or summary, it uses the title or name, which is 'Bridgy Response'

We should be better prepared for scenarios where the type is not a reply/comment and there is no summary for people who don't use the facepile. Perhpas the old comment_text we used in Semantic Linkbacks as a text fallback. "Author liked post"

@pfefferle What do you think?

pfefferle:
This is a bit more complex, now that fed.brid.gy also sends ActivityPub types, like "Follow".

quicoto:
Having the same issue, for now I need to delete those comments as they show as regular comments not "likes" 😕

khurtwilliams:
> Having the same issue, for now I need to delete those comments as they show as regular comments not "likes" 😕

> Having the same issue, for now I need to delete those comments as they show as regular comments not "likes" 😕

It might not be related to this same issue, but I see ' _Bridgy Response_' in the comments tab in the WordPress dashboard, but nothing actually displays on the blog post.

dshanske:
As a simple fallback, going to propose we display the comment type emoji instead of Bridgy response

snarfed:
Hit this now for a comment that contained a video and no text. Still an issue, I guess!

pfefferle:
@snarfed Can you share an example URL?

PR #468, opened 2024-05-02
This PR would work on both classic and fse themes

pfefferle:
related #438


pfefferle:
thanks @janboddez for your input!

dshanske (approved):
It only looks to see if protocol exists. Makes me think the code should account for ActivityPub, but since ActivityPub doesn't do display...probably not an issue.

PR #471, opened 2024-05-13
Fix #470

### LINKED ISSUE #470: Fatal error due to broken comment query callback
There's an issue here:

https://github.com/pfefferle/wordpress-webmention/blob/a9129f46da7c11876e8ff1a18ebfd90b332eaf81/includes/class-comment-walker.php#L295

This ~~alters~~ completely overwrites (but only when a `meta_query` array is present) `$query`, which is passed by reference.

And leads to a fatal error later on, when this code is run:
https://github.com/pfefferle/wordpress-webmention/blob/a9129f46da7c11876e8ff1a18ebfd90b332eaf81/includes/class-comment-walker.php#L307

Originally reported at https://github.com/janboddez/indieblocks/issues/180.

janboddez:
By reassigning `$query`, you break most that comes after.

Also, `current( $query->query_vars['meta_query'] )` _can_ return a string ... [See the example in the docs](https://developer.wordpress.org/reference/classes/wp_comment_query/#comment-2032), where `current( $query->query_vars['meta_query'] )` would return `'AND'`:
```
$args = array(
	'meta_query' => array(
		'relation' => 'AND',
		array(
			'key' => 'featured',
			'value' => '1'
		),
		array(
			'key' => 'buried',
			'value' => '1',
			'type' => 'numeric',
			'compare' => '!='
		)
	)
 );
```

But in general, `$query`, which is passed to the callback by reference, should not be reassigned.

### DIFF
--- a/includes/class-comment-walker.php
+++ b/includes/class-comment-walker.php
@@ -291,12 +291,8 @@ class Comment_Walker extends Walker_Comment {
 			return;
 		}
 
-		if ( ! empty( $query->query_vars['meta_query'] ) ) {
-			$query = current( $query->query_vars['meta_query'] );
-
-			if ( ! empty( $query['key'] ) && 'protocol' === $query['key'] ) {
-				return;
-			}
+		if ( ! empty( $query->query_vars['meta_query']['key'] ) && 'protocol' === $query->query_vars['meta_query']['key'] ) {
+			return;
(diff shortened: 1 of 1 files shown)


pfefferle:
Oh, seems that we both did that in parallel :)

PR #477, opened 2024-06-07
Speeds up the loading of posts with lots of comments

pfefferle:


dshanske:
Looks good on my phone...want to look in front of a computer. What prompted it?

pfefferle:
Ob pages with more than 100 comments I ran into a Webserver timeout on notiz.blog

dshanske:
That explains it .. I've never been that popular.

pfefferle:
Here is the example: https://notiz.blog/contact/

includes/class-comment-walker.php
```diff
@@ -288,22 +288,14 @@ public static function comment_query( $query ) {
 			return;
 		}
 
-		if ( is_admin() ) {
+		if ( is_admin() || ! is_singular() ) {
 			return;
 		}
 
-		if ( ! is_singular() ) {
+		if ( ! empty( $query->query_vars['type__in'] ) ) {
```
dshanske, line 295: Shouldn't we exclude any type specific search?  Not just in? Like type__not_in ? or even just type?

pfefferle: good point!

PR #485, opened 2024-07-22
Remove Webmention Dependence on Pings

dshanske:
This removes webmention receiving being dependent on pings_open. It does allow for checking the presence of a webmentions_closed meta key to close webmentions. If this approach is validated, will move to figure out a UI for this.

For the sending, it moves from using the pung URLs property within the post object to a new temporary storage that will just store what was sent last time, compare it to what is being send this time, and use that to send deleted webmentions.


includes/class-sender.php
```diff
@@ -208,14 +208,15 @@ public static function send_webmentions( $post_id ) {
 		$urls = webmention_extract_urls( $post->post_content, $support_media_urls );
 
 		// filter links
-		$targets = apply_filters( 'webmention_links', $urls, $post_id );
-		$targets = array_unique( $targets );
-		$pung    = get_pung( $post );
```
pfefferle, line 213: What is the problem with re-using `get_pung`? If we use something different it might result in duplicate pings. Not everyone who uses Webmentions has Pingbacks disabled.

dshanske: In the long term, I considered storing more about the ones that failed as an improvement of the retry system, but at the moment, I suppose I could revert part of this.

dshanske: @pfefferle Restored get_pung...but still have an independent webmention list. Can you have a look?

dshanske:
@pfefferle Where are we on this one?

pfefferle:
`get_pung` will now be updated, but not taken into care for Webmentions, so there might still be the possibility that we send Webmentions for URLs that were already pinged?!?

pfefferle:
I agree that we should not care about the pingback settings, but why change the `get_pung` code?

includes/functions.php
```diff
@@ -550,18 +550,23 @@ function ifset( &$var, $return = false ) { // phpcs:ignore Universal.NamingConve
...
+	if ( $post ) {
+		// Always consider the home mention link page to be open.
+		if ( get_option( 'webmention_home_mentions' ) === $post->ID ) {
+			$open = true;
+		} elseif ( ! post_type_supports( get_post_type( $post ), 'webmentions' ) ) {
+			// If the post type does not support Webmentions do not even check further.
+			$open = false;
+		} else {
+			// If the webmentions_closed meta ID has a value then consider webmentions to be open.
+			$open = get_post_meta( $post->ID, 'webmentions_closed', true );
```
pfefferle, line 565: this is very misleading! we check for `webmentions_closed` and write it into `$open`? 😳

pfefferle: I correct myself: this is not only misleading but wrong!

pfefferle:
```suggestion
			$closed = get_post_meta( $post->ID, 'webmentions_closed', true );
			$open = empty( $closed ) ? true : false;
```

pfefferle: I think the default should be `false` because we set it only when webmentions are closed?!?

includes/functions.php
```diff
@@ -550,18 +550,23 @@ function ifset( &$var, $return = false ) { // phpcs:ignore Universal.NamingConve
...
+			$open = empty( $open ) ? true : false;
```
pfefferle, line 566:
```suggestion
			$closed = (int) get_post_meta( $post->ID, 'webmentions_closed','0' );
			$open   = ! (bool) $closed;
```

pfefferle: Isn't WordPress always transforming booleans into integer-strings when using `get_post_meta`?

`true` => `'1'`
`false` => `'0'`

includes/functions.php
dshanske: If you ask to retrieve a non-existent meta key, it will return an empty string. Which I thought was the equivalent of false.

dshanske: I think I see what you mean about confusion. Trying to clean this up a bit, at the least with documentation.

includes/functions.php
dshanske: How's the amendment?

dshanske:
I like that addition. If we're happy with this... then a future addition can add the UI to close Webmentions.. but it assumes activating the plugin means you want them

pfefferle:
@dshanske this last commit adds settings to the block editor to disable Webmentions for a post.

Screenshot 2024-11-14 at 22 18 38

pfefferle:
should it be also possible to deactivate outgoing Webmentions?

dshanske:
> should it be also possible to deactivate outgoing Webmentions?

We had someone who wanted to do that on a per post basis... but I'm not sure it is a desired feature by many

pfefferle:
Ok, then we do that in a different PR!

dshanske:
I thought they got removed on purpose. Must have misread

pfefferle:
The Grunt dependencies are no longer needed... I need the other deps to work on the block-editor integration...

pfefferle:
No problem :)

dshanske:
Is this ready for merge, @pfefferle ? Want to see about some other fixes for a minor release.

pfefferle:
I think we should add some tests, to ensure that everything still works after the merge!?!

dshanske:
I'll see about getting some in and doing a dual platform check.

pfefferle (approved):

PR #491, opened 2024-09-27
Allow "Approve & Always Allow" only for Webmentions

pfefferle:
This fails on other comment types like for example ActivityPub: https://github.com/Automattic/wordpress-activitypub/issues/865


dshanske (approved):
Makes sense to me... I still want to update this feature a bit.. but it shouldn't conflict with other plugins.

Issue #474, opened 2024-05-20
Race condition causes duplicate comments for same webmention source + target


snarfed:
Hi @pfefferle, @dshanske! I got an interesting one just now: Bridgy Fed sent the same webmention (with same `source` and `target`) twice at almost the exact same time, within just ~10ms or so, and the plugin created two duplicate comments for it:

<img width="983" alt="image" src="https://github.com/pfefferle/wordpress-webmention/assets/778068/80f7bc8d-0427-4058-a115-12775c68d468">

Looks like a race condition? I'm guessing you may want to add a lock or database transaction to prevent this.

snarfed:
cc https://github.com/snarfed/bridgy-fed/issues/1063#issuecomment-2121150166

snarfed:
Still seeing this! Another example: https://snarfed.org/2024-12-01_54199#comments

PR #500, opened 2024-12-08
Finally implement async processing

pfefferle:
This should help with #474


dshanske (commented):
I don't get how this helps without enabling ASYNC by default.

pfefferle:
It helps if @snarfed enables it... but I am not sure if it should be enabled by default, because `WP_Crons` are not always properly configured...

PR #501, opened 2024-12-08
Add setting for avatar store

dshanske:
Adds a setting. Turns on the store by default.

### DIFF
--- a/includes/class-admin.php
+++ b/includes/class-admin.php
@@ -419,6 +419,16 @@ class Admin {
 		);
 		register_setting(
 			'webmention',
+			'webmention_avatar_store_enable',
+			array(
+				'type'         => 'int',
+				'description'  => esc_html__( 'Store Avatars Locally', 'webmention' ),
+				'show_in_rest' => true,
+				'default'      => 1,
+			)
+		);
+		register_setting(
+			'webmention',
 			'webmention_separate_comment',
 			array(
 				'type'         => 'int',
--- a/includes/class-admin.php
+++ b/includes/class-admin.php
@@ -424,7 +424,7 @@ class Admin {
 				'type'         => 'int',
 				'description'  => esc_html__( 'Store Avatars Locally', 'webmention' ),
 				'show_in_rest' => true,
-				'default'      => 1,
+				'default'      => 0,
 			)
 		);
 		register_setting(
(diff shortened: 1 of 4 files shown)


webmention.php
```diff
@@ -24,7 +24,7 @@
 defined( 'WEBMENTION_GRAVATAR_CACHE_TIME' ) || define( 'WEBMENTION_GRAVATAR_CACHE_TIME', WEEK_IN_SECONDS );
 
 
-defined( 'WEBMENTION_LOCAL_AVATAR_STORE' ) || define( 'WEBMENTION_LOCAL_AVATAR_STORE', false );
+defined( 'WEBMENTION_LOCAL_AVATAR_STORE' ) || define( 'WEBMENTION_LOCAL_AVATAR_STORE', true );
```
pfefferle, line 27: This seems no longer be used... so we should either remove it completely or keep it in the code.

templates/webmention-settings.php
```diff
@@ -137,6 +137,17 @@
 					</fieldset>
 				</td>
 			</tr>
+			<tr>
+				<th scope="row"><?php esc_html_e( 'Store Avatars', 'webmention' ); ?></th>
+				<td>
+					<fieldset>
+						<label for="webmention_avatar_store_enable">
+							<input type="checkbox" name="webmention_avatar_store_enable" id="webmention_avatar_store_enable" value="1" <?php echo checked( true, get_option( 'webmention_avatar_store_enable', 1 ) ); ?> />
```
pfefferle, line 145: `checked` echos by default... we should change that for the other settings too...

```suggestion
							<input type="checkbox" name="webmention_avatar_store_enable" id="webmention_avatar_store_enable" value="1" <?php checked( true, get_option( 'webmention_avatar_store_enable', 1 ) ); ?> />
```

includes/class-admin.php
```diff
@@ -417,6 +417,16 @@ public static function register_settings() {
...
 			)
 		);
+		register_setting(
+			'webmention',
+			'webmention_avatar_store_enable',
+			array(
+				'type'         => 'int',
+				'description'  => esc_html__( 'Store Avatars Locally', 'webmention' ),
+				'show_in_rest' => true,
+				'default'      => 1,
```
pfefferle, line 427: should we enable by default?

pfefferle:
This change also affects actual installations, so we would enable caching also for existing users without telling them!

We should either change it to opt-in or have the "enabled by default" only for new users!

pfefferle (approved):

Issue #527, opened 2025-07-05
Add Support for Meta Boxes If the Block Editor Isn't Used


dshanske:
https://make.wordpress.org/core/2018/11/07/meta-box-compatibility-flags/

The idea being that a traditional sidebar meta box will be added whenever the block editor isn't enabled for those individuals who use it. That covers Classic Editor and ClassicPress users, but is completely transparent for Block Editor users.

dshanske:
Addressed with #528

PR #528, opened 2025-07-05
Add classic editor fallback for webmention post settings

dshanske:
Addresses #527 

Moved the meta settings to receiver and sender functions, consistent with early design, as they aren't block specific any longer. It sets up a metabox that will appear only in the Classic Editor and Classic Press.

### DIFF
--- a/includes/class-sender.php
+++ b/includes/class-sender.php
@@ -25,6 +25,7 @@ class Sender {
 		// Send Webmentions from Every Type that Declared Webmention Support
 		$post_types = get_post_types_by_support( 'webmentions' );
 		foreach ( $post_types as $post_type ) {
+			add_action( 'save_' . $post_type, array( static::class, 'save_hook' ), 3 );
 			add_action( 'publish_' . $post_type, array( static::class, 'publish_hook' ), 3 );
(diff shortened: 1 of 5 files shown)


includes/class-sender.php
```diff
@@ -36,6 +37,41 @@ public static function init() {
 
 		// remote delete posts
 		add_action( 'webmention_delete', array( static::class, 'send_webmentions' ) );
+		\add_action( 'init', array( self::class, 'register_postmeta' ), 11 );
```
pfefferle, line 40:
```suggestion
		self::register_postmeta();
```

I think the `init` function is already triggered on init?

includes/class-sender.php
```diff
@@ -36,6 +37,41 @@ public static function init() {
...
+	}
+
+	/**
+	 * Register post meta
+	 */
+	public static function register_postmeta() {
```
pfefferle, line 46:
```suggestion
	public static function register_meta() {
```

The other function is called `register_meta` I think we should be consistent. I have no preference which one to use.

pfefferle:
Thanks @dshanske ! I have some inline feedback!

includes/class-sender.php
dshanske: Same, moved it around...fine with changing it.

PR #530, opened 2025-07-09
Fix: Implement nickname fallback for mf2 parser

pfefferle:
The current code seems to parse the `nickname`, but there is no fallback implemented. This PR tries to fall back to `nickname` and `given-name` + `family-name` if `name` is empty.

This fixes fed.brid.gy webmentions.

Screenshot 2025-07-09 at 09 33 02


pfefferle:
@snarfed fed.brid.gy is really breaking the complete classic webmention mf2 parsing!!

snarfed:
Hey @pfefferle, sorry for the trouble! Looks like the target there probably _should_ have been `https://notiz.blog/author/matthias-pfefferle/`, but we actually sent the webmention with target `https://notiz.blog/`. Evidently we got confused. Sorry about that!

pfefferle:
@snarfed any idea or advice how I can best solve or at least work around this? I get a lot of "broken" Webmentions!

PR #549, opened 2025-12-09
Add content fallback for Bridgy webmentions without text

### LINKED ISSUE #438: Bridgy Webmentions Set Content to Bridgy Response

pfefferle:
## Summary

- Extract video/audio URLs as link content when no text content property exists
- Use registered comment type excerpt templates (e.g., "liked this") as fallback for reactions (likes, reposts, etc.) that have no content

Fixes #438

## Test Plan

- [ ] Test with Bridgy video-only webmention (like https://brid.gy/comment/bluesky/...)
- [ ] Test with Bridgy like/repost webmentions
- [ ] Verify existing webmentions with text content still work correctly


dshanske (approved):
This looks good...there may be a concern about publishing links to audio/video, but that is a moderation issue, I think

PR #554, opened 2025-12-09
Fix undefined array key comment_content warning

### LINKED ISSUE #368: Undefined array key "comment_content"
If a WebMention doesn't have any content (for example a like or repost) then WordPress complains with

`PHP Warning:  Undefined array key "comment_content" in /home/blog/wp-includes/comment.php on line 2130`

Would it make sense to either add a blank space or the text "so-and-so liked this on example.com"?

pfefferle:
## Summary
- Ensure required comment data keys always exist by using `wp_parse_args` with defaults
- This prevents PHP warnings when webmentions without content (like likes or reposts) are processed

Fixes #368

## Test plan
- [ ] Receive a webmention without content (e.g., a like or repost)
- [ ] Verify no PHP warning about undefined array key `comment_content` appears


dshanske (approved):
LGTM

Issue #449, opened 2023-12-26
Add Ability to Change Avatar URL Manually or Upload to Store Manually


dshanske:

mgiannopoulos24:
Okay then. I can implement it.

PR #558, opened 2025-12-13
 Add Ability to Manually Edit and Upload Comment Avatars

mgiannopoulos24:
### Description
This PR adds the ability for administrators to manually edit the Avatar URL for a comment or upload a new avatar locally using the WordPress Media Library. Previously, this field was read-only and restricted only to comments created via the Webmention protocol.

### Related issue
Issue #449 

### Changes
- `includes/class-admin.php`:
    - Updated `enqueue_scripts` to load the WordPress Media Library (`wp_enqueue_media`) on comment editing screens.
    - Added a `save_comment` method hooked to `edit_comment` to sanitize and save the avatar meta field when the metabox is submitted.
    - `templates/webmention-edit-comment-form.php`:
    - Moved the Avatar input field outside the `if ( 'webmention' === ... )` check, making it available for all comment types.
    - Removed the `disabled` attribute to allow manual editing.
    - Added an **Upload** button and inline JavaScript to trigger the native WordPress Media Uploader modal.

### Testing Instructions
1. Navigate to **Comments** in the WordPress Admin.
2. Click **Edit** on any comment (standard comment or Webmention).
3. Locate the **Webmention Data** metabox.
4. Verify that the **Avatar** field is editable and the **Upload** button is visible.
5. Enter a custom URL or use the Upload button to select an image from the Media Library.
6. Click Update.
7. Reload the page and verify the Avatar URL has persisted.

### Showcase
- Original Comment without Avatar URL:
<p align="center">

</p>

- Comment with updated Avatar URL:
<p align="center">

</p>

<br>

<p align="center">

</p>

- Another example but with an uploaded avatar:
<p align="center">

</p>

<br>

<p align="center">

</p>

---

Let me know your thoughts on this PR.


dshanske:
The problem is it isn't uploading to the store, it is uploading to the media library, so this doesn't achieve the stated goal. You are also making it available for all comment tyhpes, which would interfere with other plugins that do avatar uploads.

The store is designed to store a unique avatar for each author URL, not for each comment URL, which makes me think.

pfefferle:
> The problem is it isn't uploading to the store, it is uploading to the media library

I agree with @dshanske here. I think we do not want to have avatars of remote users/sites in the media library.

mgiannopoulos24:
I think i fixed it now. I did some tests with 3 test users and it seemed to work. Pushing now.

mgiannopoulos24:
As requested now we have full access to the avatars through Tools panel.

We can still modify, refresh and cleanup orphan entries.

I also added the functionality to delete the avatar if we delete a user's comment. 

Of course I have taken into consideration the fact that a user may have more than 1 comments, so if we delete one of them, his avatar will still remain untouched.

dshanske:
I am going to look and figure out more specific feedback. Right now, working on a complementary idea you inspired me to do @mgiannopoulos24

dshanske:
@mgiannopoulos24 If you look at #560 this is the idea your PR made me start pursuing, which is complimentary in the sense thinking about this again caused me to ask if we were doing the right thing by overwriting the original source URL for the image in the first place. 

Assuming we merge that one, and that requires concurrence, it doesn't affect the idea of loading things into the avatar store manually, or refreshing or otherwise purging, but it is a change in where in the chain things go. Specifically it would now go Avatar Store(if enabled), followed by metadata stored URL(if present)....and then to the other fallbacks.

It means that if the store is enabled, then we should be manipulating the store, but if it isn't....

dshanske:
@mgiannopoulos24 Reading your code, adding a function to retrieve an avatar from the store directly is something we both did, and I think that should stand. Yours is probably simpler than mine, which is probably a good thing as I was considering the possibility there might be deviation in the filenames, which there shouldn't be.

Deleting orphaned avatars also makes sense, but we shoudl do this based on $host and $author properties used to generate the URL.

mgiannopoulos24:
> Deleting orphaned avatars also makes sense, but we shoudl do this based on $host and $author properties used to generate the URL.

I will modify it when I can. If you have any other suggestions let me know!

PR #560, opened 2025-12-14
Search Avatar Store Before Meta

dshanske:
This was prompted by thinking about #449 and is a change in how I'm proposing the avatar store should work.

As of now, when the store retrieves the URL, it updates the metadata with the new URL. This reverts that going forward by not storing the new local URL. Instead, it searches the store for a match, and returns it if present, but if the store is disabled, it will go back to returning the remote URL. 

That seems a lot more flexible in the long run. It also means we can focus on refreshing the store as suggested. It also means we don't lose the original URL of the image if we need it.


dshanske:
Not a bad idea, what do you think of the change overall? Do you think this is the right direction?

Will update the PR.

dshanske:
This should address the feedback you gave.

pfefferle (approved):

